PR #67, opened 2026-01-19
feat: add non-interactive mode to create command

### LINKED ISSUE #66: Non interactive create command
# Summary

Add option to run the create command with non interactive flow - using flags and so on

kfirstri:
Add support for running the create command with flags for non-interactive use:
- --template/-t: Specify template to use
- --name/-n: Project name
- --description/-d: Project description (optional)
- --path/-p: Where to create the project (optional)
- --push-entities/--no-push-entities: Control entity pushing
- --deploy/--no-deploy: Control site deployment

The command now works in both interactive and non-interactive modes.

Closes #66

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/cli/commands/project/create.ts
+++ b/src/cli/commands/project/create.ts
@@ -14,7 +14,7 @@ import type { RunCommandResult } from "../../utils/runCommand.js";
 
 const orange = chalk.hex("#E86B3C");
 const cyan = chalk.hex("#00D4FF");
-const DEFAULT_TEMPLATE_NAME = "backend-only";
+const DEFAULT_TEMPLATE_ID = "backend-only";
 
 interface CreateOptions {
   name?: string;
@@ -25,9 +25,9 @@ interface CreateOptions {
 
 async function getDefaultTemplate(): Promise<Template> {
   const templates = await listTemplates();
-  const template = templates.find((t) => t.name === DEFAULT_TEMPLATE_NAME);
+  const template = templates.find((t) => t.id === DEFAULT_TEMPLATE_ID);
   if (!template) {
-    throw new Error(`Default template "${DEFAULT_TEMPLATE_NAME}" not found`);
+    throw new Error(`Default template "${DEFAULT_TEMPLATE_ID}" not found`);
   }
   return template;
 }


src/cli/commands/project/create.ts
```diff
@@ -151,6 +209,14 @@ async function create(): Promise<RunCommandResult> {
...
   .description("Create a new Base44 project")
-  .action(async () => {
-    await runCommand(create, { fullBanner: true, requireAuth: true });
+  .option("-t, --template <template>", "Template to use")
+  .option("-n, --name <name>", "Project name")
+  .option("-d, --description <description>", "Project description")
+  .option("-p, --path <path>", "Path where to create the project")
+  .option("--push-entities", "Automatically push entities after creation")
+  .option("--no-push-entities", "Skip pushing entities")
+  .option("--deploy", "Automatically deploy the site after creation")
```
kfirstri: @claude lets make it so only if the --deploy flag is specified then we go through the deployment -> build, push entities and site deployed. if it's not there don't do it.

So the only flags you should have-
* -t --template (required)
* -n --name (rquired)
* -d --description (optional)
* -p --path (optional)
* --deploy

src/cli/commands/project/create.ts
```diff
@@ -151,6 +208,11 @@ async function create(): Promise<RunCommandResult> {
 
 export const createCommand = new Command("create")
   .description("Create a new Base44 project")
-  .action(async () => {
-    await runCommand(create, { fullBanner: true, requireAuth: true });
+  .option("-t, --template <template>", "Template to use")
```
kfirstri: @claude lets remove this option, i always want to use the "backend only" template for now

src/cli/commands/project/create.ts
```diff
@@ -151,6 +208,11 @@ async function create(): Promise<RunCommandResult> {
...
+  .option("-n, --name <name>", "Project name")
+  .option("-d, --description <description>", "Project description")
+  .option("-p, --path <path>", "Path where to create the project")
+  .option("--deploy", "Build and deploy the site (includes pushing entities)")
+  .action(async (options: CreateOptions) => {
+    await runCommand(() => create(options), { fullBanner: true, requireAuth: true });
```
kfirstri: can we make the fullBanner detect if we're in a non interactive environment and change this to false? @claude

src/cli/commands/project/create.ts
```diff
@@ -15,49 +15,88 @@ import type { RunCommandResult } from "../../utils/runCommand.js";
...
-    },
-    {
-      onCancel: onPromptCancel,
+  // Use provided options or prompt interactively
+  let template: Template;
+  let name: string;
+  let description: string | undefined;
+  let projectPath: string;
+
+  if (options.template && options.name) {
```
kfirstri: @claude lets detect non interactiveness before and split the code to two function in this files to be more readable

PR #110, opened 2026-01-25
feat: Auto-open browser for device code verification in login

gonengar:
## Summary

- Add "Press Enter to open in browser..." prompt to auth login flow for better UX
- Auto-open the verification URL in the user's default browser when they press Enter
- Style the verification URL with `theme.colors.links()` for visual consistency

## Test plan

- [ ] Run `base44 auth login` command
- [ ] Verify the verification code and URL are displayed (URL should be cyan colored)
- [ ] Verify "Press Enter to open in browser..." prompt appears
- [ ] Press Enter and confirm browser opens to the verification URL
- [ ] Complete authentication and verify login succeeds

### DIFF
--- a/src/cli/commands/auth/login.ts
+++ b/src/cli/commands/auth/login.ts
@@ -2,7 +2,6 @@ import { Command } from "commander";
 import { log } from "@clack/prompts";
 import pWaitFor from "p-wait-for";
 import open from "open";
-import * as readline from "readline";
 import {
   writeAuth,
   generateDeviceCode,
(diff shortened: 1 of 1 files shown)


src/cli/commands/auth/login.ts
```diff
@@ -30,9 +32,23 @@ async function generateAndDisplayDeviceCode(): Promise<DeviceCodeResponse> {
...
+    `\nPlease confirm this code at: ${theme.colors.links(deviceCodeResponse.verificationUri)}`
   );
 
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  await new Promise<void>((resolve) => {
+    rl.question("Press Enter to open in browser...", () => {
```
kfirstri: we don't want to use readline in the project, we cause use clack prompts to show something like this, i would also claim that maybe we don't want the even ask the user? we already have `open` installed (for the `dashboard` command), lets just open the browser? we can also add a check before if we're is a non-tty environment and not open the browser.
@claude suggest a fix please

PR #112, opened 2026-01-25
agents integration

Paveltarno:
## Description

This PR implements a complete agents integration system for the CLI, enabling users to define AI agents with custom instructions and tool configurations. The system provides push/pull commands to synchronize agent configurations between local JSON/JSONC files and Base44, includes comprehensive schema validation with Zod, and follows the established resource pattern used by entities and functions.

## Related Issue

None

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Other (please describe):

## Changes Made

- Added complete agent resource abstraction layer in `src/core/resources/agent/` with schema validation, config reading/writing, and API integration
- Implemented `agents push` command that replaces all remote agent configurations on Base44
- Implemented `agents pull` command that fetches agents from Base44 and replaces all local agent files
- Created agent configuration schema supporting entity tools (with granular CRUD permissions) and backend function tools
- Integrated agent loading into project config reader with duplicate name detection
- Added `agentsDir` field to project config schema (defaults to "agents")
- Implemented API client methods `pushAgents` and `fetchAgents` for Base44 synchronization
- Created comprehensive test fixtures for valid agents, invalid schemas, and duplicate names
- Added unit tests covering agent reading, validation, API sync, and error cases (140+ lines of tests)
- Extracted API error handling utilities (`formatApiError`, `ApiErrorResponse`) to base44-client module for reuse
- Updated AGENTS.md documentation to reflect new agent resource architecture
- Registered agents commands in CLI program
- Added consistent outro messages to push commands across resources (entities, functions, agents)

## Testing

- [x] I have tested these changes locally
- [x] I have added/updated tests as needed
- [x] All tests pass (`npm test`)

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings
- [x] I have updated AGENTS.md if I made architectural changes

## Additional Notes

**Agent Configuration Format:**
- Agent names must be lowercase alphanumeric with underscores only (validated via regex: `^[a-z0-9_]+$`)
- Entity tool configs support granular operation permissions: read, create, update, delete
- Backend function tool configs include custom descriptions for tool use
- Optional WhatsApp greeting messages for conversational interfaces
- Both JSON and JSONC file formats supported with comment preservation

**Architecture Notes:**
- The push/pull commands use a full replacement strategy (not incremental sync) - pushing replaces all remote configs, pulling replaces all local files
- This ensures deterministic state synchronization and follows the same pattern as the entities resource
- Includes proper error handling for API failures and validation errors with clear user feedback
- API error formatting utilities refactored to base44-client module for better code organization and reuse
- Uses `looseObject` schema for API flexibility while still validating required fields like agent name format

---
<sub>🤖 Generated by Claude | 2026-01-27 14:00 UTC</sub>

### DIFF
--- a/src/core/resources/agent/schema.ts
+++ b/src/core/resources/agent/schema.ts
@@ -18,15 +18,10 @@ export const AgentConfigApiResponseSchema = z.looseObject({
   name: z.string(),
 });
 
-export const ListAgentsResponseSchema = z
-  .object({
-    items: z.array(AgentConfigApiResponseSchema),
-    total: z.number(),
-  })
-  .transform((data) => ({
-    items: data.items,
-    total: data.total,
-  }));
+export const ListAgentsResponseSchema = z.object({
+  items: z.array(AgentConfigApiResponseSchema),
+  total: z.number(),
+});
 
 export type AgentConfigApiResponse = z.infer<typeof AgentConfigApiResponseSchema>;
 export type ListAgentsResponse = z.infer<typeof ListAgentsResponseSchema>;


src/core/clients/base44-client.ts
```diff
@@ -13,6 +13,27 @@ import {
 } from "@/core/auth/config.js";
 import { getAppConfig } from "@/core/project/index.js";
 
+/**
+ * Standard API error response format from the Base44 backend.
+ * All HTTP errors from FastAPI are returned in this format.
+ */
+export interface ApiErrorResponse {
```
kfirstri: Lets move this to clients/schems.ts with a ApiErrorSchema and use optional / parse to stick to the structure of the project

src/core/clients/base44-client.ts
```diff
@@ -13,6 +13,27 @@ import {
...
+  message: string;
+  detail: string | Record<string, unknown> | unknown[];
+  traceback?: string;
+}
+
+/**
+ * Formats an API error response into a human-readable string.
+ * Prefers `message` (human-readable) over `detail`.
+ */
+export function formatApiError(errorJson: unknown): string {
```
kfirstri, line 17: Can we try and catch and identify errors inside the ky client itself? and then we don't need to export this function and use it evereywhere - the base client will catch errors with this structure and throw correctly.. wdyt?
I'de say it's probably something to do in a different PR though

src/cli/commands/agents/push.ts
```diff
@@ -0,0 +1,51 @@
...
+    log.success(`Updated: ${result.updated.join(", ")}`);
+  }
+  if (result.deleted.length > 0) {
+    log.warn(`Deleted: ${result.deleted.join(", ")}`);
+  }
+
+  return {};
+}
+
+export const agentsCommand = new Command("agents")
```
kfirstri: Lets move the new "agents" command to be under commands/agents/index.ts, and each file will just export it's command

src/core/resources/agent/schema.ts
kfirstri, line 1: In general - do we need all of the schema?
We're trying to just zod-ify only things we actually use inside the CLI codebase, we don't want to verify the entire schema right now. might worth adding to AGENTS.md as well

kfirstri: Also, do all the agent.json file as written as snake_case and not camelCase?

src/core/resources/agent/schema.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type SyncAgentsResponse = z.infer<typeof SyncAgentsResponseSchema>;
+
+export const AgentConfigApiResponseSchema = z.object({
+  name: z.string(),
+  description: z.string(),
+  instructions: z.string(),
+  tool_configs: z.array(ToolConfigSchema).default([]),
+  whatsapp_greeting: z.string().nullable().optional(),
+});
```
kfirstri, line 19: You can look at other resources schema, lets use .transform to change response into camelCase

src/core/resources/agent/config.ts
```diff
@@ -0,0 +1,76 @@
...
+  for (const agent of agents) {
+    const filePath = join(agentsDir, `${agent.name}.${CONFIG_FILE_EXTENSION}`);
+    await writeJsonFile(filePath, {
+      name: agent.name,
+      description: agent.description,
+      instructions: agent.instructions,
+      tool_configs: agent.tool_configs,
+      whatsapp_greeting: agent.whatsapp_greeting ?? null,
+    });
+  }
```
kfirstri: I wonder if we should use an ejs file for agent.json file?
The main benefit is that we can give the user a file with comments inside in case they want to edit it..

we do something similar with .app.json file
```
export function generateAppConfigContent(id: string): string {
  return `// Base44 App Configuration
// This file links your local project to your Base44 app.
// Do not commit this file to version control.
{
  "id": "${id}"
}
`;
}
```
we can have another function with corrent agents structure, but also with more context about each field.. wdyt?

src/core/resources/agent/config.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  return agents;
+}
+
+export async function writeAgents(
+  agentsDir: string,
+  agents: AgentConfigApiResponse[]
+): Promise<{ written: string[]; deleted: string[] }> {
+  const existingAgents = await readAllAgents(agentsDir);
```
kfirstri, line 61: I think agents is something we should get as an argument, in the command that calls this function we already do getProjectConfig which reads all the agents anyway, so lets pass it to this function instead of reading it again

src/core/resources/agent/config.ts
```diff
@@ -0,0 +1,76 @@
...
+  const newNames = new Set(agents.map((a) => a.name));
+
+  const toDelete = existingAgents.filter((a) => !newNames.has(a.name));
+  for (const agent of toDelete) {
+    const filePath = join(agentsDir, `${agent.name}.${CONFIG_FILE_EXTENSION}`);
```
kfirstri: NIT: might be a bug if the AI / user will start creating agent config files using JSON and not JSONC, a bit of an edge case but it might be worth to globby and check for either endings >.>

src/core/clients/base44-client.ts
Paveltarno: yes good idea! as long as all errors are returned the same way from the API what you are suggesting is the correct approach

src/core/clients/base44-client.ts
Paveltarno: yes been contemplating on where to put this

src/core/resources/agent/config.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { AgentConfig, AgentConfigApiResponse } from "./schema.js";
+import { CONFIG_FILE_EXTENSION_GLOB, CONFIG_FILE_EXTENSION } from "../../consts.js";
+
+async function readAgentFile(agentPath: string): Promise<AgentConfig> {
+  const parsed = await readJsonFile(agentPath);
+  const result = AgentConfigSchema.safeParse(parsed);
+
+  if (!result.success) {
+    throw new Error(
+      `Invalid agent configuration in ${agentPath}: ${result.error.issues
```
kfirstri: I think this example needs to be fixed in AGENTS.md file,
lets change to just use .parse in line 10 and let it throw, if alternative is to just use `results.error.message`

src/cli/commands/agents/push.ts
```diff
@@ -0,0 +1,51 @@
...
+import { agentsPullCommand } from "./pull.js";
+
+async function pushAgentsAction(): Promise<RunCommandResult> {
+  const { agents } = await readProjectConfig();
+
+  log.info(
+    agents.length === 0
+      ? "No local agents found - this will delete all remote agents"
+      : `Found ${agents.length} agents to push`
+  );
```
kfirstri, line 15: Do we want to prompt the user in case of agents.length===0? we're writing "this will delete all remote agents" and then just going with it. how about a prompt and a --yes/-y flag?

src/core/resources/agent/schema.ts
Paveltarno: so you are suggesting we type it without zod and leave the "restrictions" to agents.md?

src/cli/commands/agents/pull.ts
```diff
@@ -0,0 +1,55 @@
...
+  );
+
+  if (response.items.length === 0) {
+    return { outroMessage: "No agents found on Base44" };
+  }
+
+  const { written, deleted } = await runTask(
+    "Writing agent files",
+    async () => {
+      return await writeAgents(agentsDir, response.items);
```
kfirstri: instead of passing the agentsDir, on line 10 you can do 
```
const { project, agents } = await readProjectConfig();
```
and just pass the list of agent to the writeAgents function

src/cli/commands/agents/pull.ts
```diff
@@ -0,0 +1,55 @@
...
+import { runCommand, runTask } from "../../utils/index.js";
+import type { RunCommandResult } from "../../utils/runCommand.js";
+
+async function pullAgentsAction(): Promise<RunCommandResult> {
+  const { project } = await readProjectConfig();
+
+  const configDir = dirname(project.configPath);
+  const agentsDir = join(configDir, project.agentsDir);
+
+  const response = await runTask(
```
kfirstri: how about call this `remoteAgents`?

src/cli/commands/agents/push.ts
```diff
@@ -0,0 +1,51 @@
...
+    log.success(`Created: ${result.created.join(", ")}`);
+  }
+  if (result.updated.length > 0) {
+    log.success(`Updated: ${result.updated.join(", ")}`);
+  }
+  if (result.deleted.length > 0) {
+    log.warn(`Deleted: ${result.deleted.join(", ")}`);
+  }
+
+  return {};
```
kfirstri: i saw that the pull command has an outroMessage, but the push command doesn't, lets align

src/cli/commands/agents/push.ts
Paveltarno: hmm but didn't we say that for now we do not interrupt the push/pull in any way? because this is just an edge case in this whole conflict scenario
i suggest we either fully tackle it not at all
wdyt?

src/core/clients/base44-client.ts
kfirstri: so we can either do it here, or resolve this in another pr and remove from this one..
Let me know what you choose and we i'll open an issue if needed

src/core/resources/agent/schema.ts
kfirstri: No - I want to use zod, but only type the field that we need for CLI usage, like name and stuff..
I think we will need anyway to release jsonSchemas for all of our files and this will help the AI more then CLI errors

src/core/resources/agent/config.ts
Paveltarno: sure np, commented file is a good thing

src/core/clients/base44-client.ts
Paveltarno: let's do it in another PR, @claude open an issue 😬

src/core/resources/agent/config.ts
Paveltarno: good idea

src/cli/commands/agents/push.ts
kfirstri: fair 👍

src/cli/commands/agents/pull.ts
Paveltarno: hmm not sure i follow,readProjectConfig returns the localAgents while in this fn we pull the remote agents and override the local files

kfirstri: mm you're correct, so keep passing what you pass already, mainly i just see that in writeAgents you do 
`const existingAgents = await readAllAgents(agentsDir);`
so maybe this you can also pass into the function instead of reading it again?

kfirstri: btw another idea, a pattern i did for function (it depends what kind of flow you want to do) - 
1. I create in core/function/deploy.ts
2. this file exposes one function calls "deploy", but it does 2 things - read relevant files + call the api method
Just thinking that you can do some similar pattern - a "pull.ts" file that calls the api, resolves, and write to folder - and then you can just call it from the command, no need to "orchestrate" it inside the command.. whatever you like

src/core/resources/agent/config.ts
Paveltarno: played with it a bit i think for now let's leave the same as generateAppConfig (without an ejs file) and then maybe align the whole system

src/cli/commands/agents/push.ts
Paveltarno: kk, will do that (better to align push on entites/functions as well)

src/core/resources/agent/config.ts
Paveltarno: fixing, but what example in agents.md exactly? couldn't find anything applicable

Paveltarno: i've also fixed entity to align

src/core/resources/agent/schema.ts
Paveltarno: ok i've made it more loose
in the end of the day i think the server should publish it's apis in swagger or something and we'll construct everything from that (json-schemas, zods, etc)

src/core/resources/agent/schema.ts
Paveltarno: why do we need camelCase though? don't we transfer this to the config file as is? (config file is also snake_case)

src/core/clients/base44-client.ts
Paveltarno: @claude  ok so just open a PR, not an issue

PR #249, opened 2026-02-16
CLI local dev functions

artemdemo:
> [!NOTE]
> ## Description
>
> This PR adds local development support for Base44 backend functions by introducing a Deno runtime integration. When running `base44 dev`, functions are automatically started in Deno subprocesses and proxied through the dev server, allowing developers to test function deployments locally without deploying to production.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Created `deno-runtime/` directory with a Deno-specific wrapper that patches `Deno.serve` to inject dynamic ports
> - Added `FunctionManager` class to spawn, manage, and proxy Deno function processes with port allocation and lifecycle management
> - Integrated function routing into dev server at `/api/apps/:appId/functions/:functionName`
> - Added `DependencyNotFoundError` to handle missing Deno installation gracefully with helpful error messages
> - Created `createDevLogger` utility for consistent dev server logging with color-coded output
> - Updated build pipeline to bundle the Deno runtime wrapper alongside the CLI
> - Added separate `tsconfig.json` for Deno runtime to use `@types/deno` instead of Node types, preventing type conflicts
> - Verify Deno installation at dev server startup and provide clear error messages if missing
> - Updated `biome.json` to exclude entire `dist/` and `node_modules/` directories
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> - The Deno wrapper uses environment variables (`FUNCTION_PATH`, `FUNCTION_PORT`, `FUNCTION_NAME`) to configure function execution
> - Functions are started on-demand when first accessed and remain running until dev server shutdown
> - The `FunctionManager` handles process lifecycle, stdout/stderr piping, and graceful shutdown on SIGINT/SIGTERM
> - Updated `docs/error-handling.md` to document the new `DEPENDENCY_NOT_FOUND` error code
> - Added README.md in `deno-runtime/` explaining the separation between Node.js CLI and Deno function runtime
>
> ---
> <sub>🤖 Generated by Claude | 2026-02-18 16:45 UTC</sub>

### DIFF
--- a/src/cli/dev/dev-server/function-manager.ts
+++ b/src/cli/dev/dev-server/function-manager.ts
@@ -33,7 +33,9 @@ export class FunctionManager {
     this.functions = new Map(functions.map((f) => [f.name, f]));
     this.logger = logger;
 
-    this.verifyDenoIsInstalled();
+    if (functions.length > 0) {
+      this.verifyDenoIsInstalled();
+    }
   }
 
   private verifyDenoIsInstalled(): void {
--- a/src/cli/dev/dev-server/main.ts
+++ b/src/cli/dev/dev-server/main.ts
@@ -63,10 +63,10 @@ export async function createDevServer(
     clackLog.info(
       `Loaded functions: ${functionManager.getFunctionNames().join(", ")}`,
     );
-  }
 
-  const functionRoutes = createFunctionRouter(functionManager, devLogger);
-  app.use("/api/apps/:appId/functions", functionRoutes);
+    const functionRoutes = createFunctionRouter(functionManager, devLogger);
+    app.use("/api/apps/:appId/functions", functionRoutes);
+  }
 
   app.use((req, res, next) => {
     return remoteProxy(req, res, next);


deno-runtime/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../tsconfig.base.json",
+  "compilerOptions": {
+    "lib": ["ES2022", "DOM"],
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "typeRoots": ["../node_modules/@types"],
+    "types": ["deno"]
```
artemdemo, line 8: Deno types are different from Node and in order to handle it without collision between types I need separate `tsconfig`

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,216 @@
...
+import getPort from "get-port";
+import {
+  DependencyNotFoundError,
+  InternalError,
+  InvalidInputError,
+} from "@/core/errors.js";
+import type { BackendFunction } from "@/core/resources/function/schema.js";
+import type { Logger } from "../createDevLogger";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
```
artemdemo, line 14: This is the way to figure out where will be bundled deno runtime wrapper that I use for functions.

`import.meta.url` will be something like `file:///Users/username/Development/test-project/cli/dist/cli/index.js`

and I know that output of the wrapper will be next to `cli` folder

src/cli/dev/dev-server/main.ts
```diff
@@ -46,6 +57,24 @@ export async function createDevServer(
...
+
+  const devLogger = createDevLogger(false);
+
+  const functionManager = new FunctionManager(functions, devLogger);
+  functionManager.verifyDenoIsInstalled();
+
+  if (functionManager.functionNames().length > 0) {
+    clackLog.info(
+      `Loaded functions: ${functionManager.functionNames().join(", ")}`,
+    );
```
artemdemo, line 65: here I'm using `clack.log` because server is not ready yet and `dev` command didn't send `outro` of `clack`

biome.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.3.13/schema.json",
```
artemdemo, line 2: no related to PR, but now schema is aligned to our version

biome.json
```diff
@@ -13,7 +13,7 @@
     }
   },
   "files": {
-    "includes": ["**", "!**/dist/**", "!**/node_modules/**", "!**/tests/fixtures/**", "!**/*.d.ts"]
```
artemdemo, line 16: no related to PR, but the correct syntaxis is not to mention `**` at the end

artemdemo:
@claude review

artemdemo:
@claude review

src/cli/dev/createDevLogger.ts
```diff
@@ -0,0 +1,66 @@
...
+  hour12: false,
+});
+
+const colorByType: Record<LogType, (text: string) => string> = {
+  error: theme.styles.error,
+  warn: theme.styles.warn,
+  log: (text: string) => text,
+};
+
+export function createDevLogger(isPrefixed = true): Logger {
```
kfirstri: I got a question about this in general - 
I see today we only intialize this in 1 place (`dev-server/main.ts`) as `createDevLogger(false)`..

1. so if i understand correctly the non-prefixed logger is for different messages like "function started" and stuff.. an eventually logs from the function will have the time prefix?

src/cli/dev/createDevLogger.ts
```diff
@@ -0,0 +1,66 @@
...
+    switch (type) {
+      case "error":
+        console.error(colorize(msg));
+        break;
+      case "warn":
+        console.warn(colorize(msg));
+        break;
+      default:
+        console.log(msg);
+    }
```
kfirstri: technically i think we can resolve the switch also from an object with key as type, and value as the correct function to use

src/cli/dev/createDevLogger.ts
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  const prefixedLog = (type: LogType, msg: string) => {
+    const timestamp = dateTimeFormat.format(new Date());
+    const colorize = colorByType[type];
+    console.log(`${theme.styles.dim(timestamp)} ${colorize(msg)}`);
```
kfirstri: Not sure i understand why in the `print` we use the correct console.<Something> function, but when prefixed we always use console.log

src/cli/dev/createDevLogger.ts
kfirstri: How about we consider something like https://github.com/fullstack-build/tslog tslog or some logging framework which covers all our cases?

src/cli/dev/dev-server/main.ts
```diff
@@ -17,9 +24,13 @@ interface DevServerResult {
...
 
 export async function createDevServer(
-  options: DevServerOptions = {},
+  options: DevServerOptions,
 ): Promise<DevServerResult> {
-  const port = options.port ?? (await getPort({ port: DEFAULT_PORT }));
+  const { port: userPort } = options;
+  const port = userPort ?? (await getPort({ port: DEFAULT_PORT }));
+
+  const { project } = await readProjectConfig();
```
kfirstri: What do you think about injecting the project config object, and the function data as an argument to the `createDevServer` function? it will help us test this more easily..

1. you can grab the project / function at the `dev.ts` command
2. pass it to the createDevServer
3. less coupling for the dev-server 💪 


sharing some thoughts: ideally we would have a `coreSDK` that would've been passed into `createDevServer`, and then when you'll do stuff like `const project = coreSDK.project` or `const functions = coreSDK.functions.getAll()` we would be able to mock the SDK in tests..

src/cli/dev/dev-server/routes/functions.ts
```diff
@@ -0,0 +1,107 @@
+import type { IncomingMessage } from "node:http";
+import { request as httpRequest } from "node:http";
+import type { Request, Response } from "express";
+import { Router } from "express";
+import type { Logger } from "../../createDevLogger.js";
+import type { FunctionManager } from "../function-manager.js";
+
+export function createFunctionRoutes(
```
kfirstri: nit: createFunctionRouter?

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+  private running: Map<string, RunningFunction> = new Map();
+  private starting: Map<string, Promise<number>> = new Map();
+  private logger: Logger;
+
+  constructor(functions: BackendFunction[], logger: Logger) {
+    this.functions = new Map(functions.map((f) => [f.name, f]));
+    this.logger = logger;
+  }
+
+  functionNames(): string[] {
```
kfirstri: nit: getFunctionNames() ? really not critical

src/cli/dev/dev-server/routes/functions.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Proxy an Express request to a Deno function running on the specified port.
+ * Forwards headers, body, and method.
+ */
+function proxyRequest(
+  req: Request,
+  res: Response,
+  port: number,
+  logger: Logger,
+): Promise<void> {
```
kfirstri: Since we're using `http-proxy-middleware` we already have `http-proxy` bundled, we can add it as a dependency and just use here instead of implementing this by ourselves.
Im conflicted about this because this implementation is pretty straightforward, maybe http-proxy is an overkill.

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+    return Array.from(this.functions.keys());
+  }
+
+  getFunction(name: string): BackendFunction | undefined {
+    return this.functions.get(name);
+  }
+
+  verifyDenoIsInstalled(): void {
+    if (this.functions.size > 0) {
```
kfirstri: How about we move this verification to be part of the constructor, and only create a new `FunctionManager` if there are any functions?

in main.ts:
```ts

if (functions.length > 0) {
  const functionManager = new FunctionManager(functions, devLogger);

  const functionRoutes = createFunctionRoutes(functionManager, devLogger);
  app.use("/api/apps/:appId/functions", functionRoutes);

  clackLog.info(
      `Loaded functions: ${functionManager.functionNames().join(", ")}`,
  );
}
```

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (pending) {
+      return pending;
+    }
+
+    const backendFunction = this.functions.get(name);
+    if (!backendFunction) {
+      throw new InvalidInputError(`Function "${name}" not found`, {
+        hints: [{ message: "Check available functions in your project" }],
+      });
+    }
```
kfirstri: nit: my instinct is this should be the first check, but doing .get from a Map is cheap so not critical

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+      ready: false,
+    };
+
+    this.running.set(name, runningFunc);
+    this.setupProcessHandlers(name, process);
+
+    return this.waitForReady(name, runningFunc);
+  }
+
+  getPort(name: string): number | undefined {
```
kfirstri: not used anywhere.. and also i guess the right way to do this is to call `ensureRunning`?

kfirstri: and btw, can you check knip.json configuration to make sure it's scanning your stuff? i would assume knip to catch this

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    return process;
+  }
+
+  private setupProcessHandlers(name: string, process: ChildProcess): void {
+    // Pipe stdout with function name prefix
+    process.stdout?.on("data", (data: Buffer) => {
+      const lines = data.toString().trim().split("\n");
+      for (const line of lines) {
+        this.logger.log(line);
```
kfirstri, line 138: since we passed a non-prefixed logger to the FunctionManager these will be without a timestamp, what's our design here? we want the timestamp? maybe we need 2 different loggers?

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+                { message: "Check the function code for startup errors" },
+              ],
+            },
+          ),
+        );
+      }, READY_TIMEOUT);
+
+      const onData = (data: Buffer) => {
+        const output = data.toString();
+        if (output.includes("Listening on")) {
```
kfirstri, line 196: we have a comment in the `deno-runtime/main.ts` about the "listening on" string (_// This message is used by FunctionManager to detect when the function is ready_), lets add a comment here just so it will be clear that this is a contract

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -0,0 +1,222 @@
...
+  constructor(functions: BackendFunction[], logger: Logger) {
+    this.functions = new Map(functions.map((f) => [f.name, f]));
+    this.logger = logger;
+  }
+
+  functionNames(): string[] {
+    return Array.from(this.functions.keys());
+  }
+
+  getFunction(name: string): BackendFunction | undefined {
```
kfirstri: I think it's good for now, but something is weird to me with the `FunctionManager` API, specifically `getFunction` that just returns the BackendFunction type, but not a status.. basically i don't have a way to check if a function is running, only to `ensureRunning`. 

In the `createFunctionRoutes` we first call `getFunction` to check it exists and then `ensureRunning` to start and get the port.. but technically we could've just skipped the `getFunction` since if the function didn't exists, the `ensureRunning` would've thrown `InvalidInputError`..
Another way is to do `manager.functionNames().includes(name)` to check if a function exists..

just thoughts, we can talk about this more

src/cli/dev/dev-server/routes/functions.ts
artemdemo: ok

src/cli/dev/dev-server/function-manager.ts
artemdemo: ok

src/cli/dev/createDevLogger.ts
artemdemo: good point. Also I don't really use `prefixedLog` (the fact that every log is prefixed with timestamp makes it harder to copy), so I will remove it, max we will add it later if needed.

src/cli/dev/createDevLogger.ts
artemdemo: correct

src/cli/dev/dev-server/function-manager.ts
artemdemo: but part of which constructor?
I'm creating `FunctionManager` in `main.ts` in `createDevServer()`

src/cli/dev/dev-server/routes/functions.ts
artemdemo: ok

src/cli/dev/dev-server/function-manager.ts
artemdemo: moving

src/cli/dev/dev-server/function-manager.ts
artemdemo: you're right, removing. `ensureRunning` is enough

src/cli/dev/dev-server/function-manager.ts
artemdemo: 1. removing
2. `knip` config lacks property for class methods, so it doesn't check it. Will add it in a separate PR.

src/cli/dev/dev-server/function-manager.ts
artemdemo: I thought about it - and I prefer not to use prefix with timestamps - it makes logs harder to copy from CLI.
So I will remove logic that adds timestamps.
We will add it later if we will want to do so.

src/cli/dev/dev-server/function-manager.ts
artemdemo: ok

src/cli/dev/createDevLogger.ts
artemdemo: removing prefix for now, will add later if needed

artemdemo: no, I didn't but it looks like an overkill, I think it's better to keep it simple for now, since we just decided to drop Ink.
We will see how it goes and will add complexity step by step

src/cli/dev/dev-server/function-manager.ts
artemdemo: https://github.com/base44/cli/pull/263

src/cli/dev/dev-server/function-manager.ts
kfirstri: I suggest moving the verification to the FunctionManager constructure, and only initialize a FunctionManager if there are any functions

artemdemo: Will not work, since I'm doing garbage collection if dev server is going down.
In other words I'm calling `functionManager.stopAll();` on line 95 in `src/cli/dev/dev-server/main.ts`

But I did some reorg so it will look better - moved `verifyDenoIsInstalled` call into `FunctionManager`

src/cli/dev/dev-server/main.ts
artemdemo: sure, moved

PR #259, opened 2026-02-17
feat: add entity records CRUD commands

ayal:
> [!NOTE]
> ## Description
>
> This PR adds full CRUD support for entity records via a new `entities records` subcommand group. Users can now list, get, create, update, and delete records for any entity directly from the CLI, authenticated via a token exchange from the platform token to an app-user token with RLS bypass headers for admin-level access.
>
> The `entities push` command was also refactored: previously it was registered as the top-level `entities` command; it is now a proper subcommand under a new `entities` parent, enabling the `records` subcommand to be added alongside it.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `entities records list` command with `--filter`, `--sort`, `--limit`, `--skip`, and `--fields` options
> - Added `entities records get` command to fetch a single record by ID
> - Added `entities records create` command accepting record data via `--data` (inline JSON/JSONC) or `--file`
> - Added `entities records update` command accepting update fields via `--data` or `--file`
> - Added `entities records delete` command with an interactive confirmation prompt and `--yes` flag to skip it
> - Added `getAppUserClient()` in `base44-client.ts` that exchanges the platform token for an app-user token and attaches `X-Bypass-RLS` / `X-Bypass-Entities-Filter-Limit` headers
> - Added `records-api.ts` with core `listRecords`, `getRecord`, `createRecord`, `updateRecord`, `deleteRecord` functions with Zod validation on all responses
> - Added `records-schema.ts` with `EntityRecordSchema` (loose object) and `DeleteRecordResponseSchema`
> - Refactored `entities push` from a top-level command into a subcommand under a new `entities` parent command group
> - Extended `Base44APIMock` with helpers for mocking all record endpoints and the token-exchange endpoint
> - Added `toContainInStdout` assertion helper to `CLIResultMatcher`
> - Updated README command table with the five new `entities records` commands
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `delete` command uses `CLIExitError(0)` for a clean exit when the user cancels the confirmation prompt, consistent with the project's error handling patterns. The `EntityRecordSchema` uses `z.looseObject` to allow arbitrary dynamic data fields beyond the required system fields (`id`, `created_date`, `updated_date`, `created_by`).
>
> ---
> <sub>🤖 Generated by Claude | 2026-02-23 00:00 UTC</sub>

### DIFF
--- a/src/core/clients/index.ts
+++ b/src/core/clients/index.ts
@@ -1,2 +1,6 @@
-export { base44Client, getAppClient, getAppUserClient } from "./base44-client.js";
+export {
+  base44Client,
+  getAppClient,
+  getAppUserClient,
+} from "./base44-client.js";
 export { oauthClient } from "./oauth-client.js";
--- a/tests/cli/testkit/Base44APIMock.ts
+++ b/tests/cli/testkit/Base44APIMock.ts
@@ -280,9 +280,8 @@ export class Base44APIMock {
     if (this.appUserTokenMocked) return this;
     this.appUserTokenMocked = true;
     this.handlers.push(
-      http.get(
-        `${BASE_URL}/api/apps/${this.appId}/auth/token`,
-        () => HttpResponse.json({ token: "mock-app-user-token" }),
+      http.get(`${BASE_URL}/api/apps/${this.appId}/auth/token`, () =>
+        HttpResponse.json({ token: "mock-app-user-token" }),
       ),
     );
     return this;


src/core/resources/entity/records-schema.ts
```diff
@@ -0,0 +1,26 @@
...
+ * System fields (id, created_date, updated_date, created_by) are always present.
+ */
+export const EntityRecordSchema = z
+  .object({
+    id: z.string(),
+    created_date: z.string(),
+    updated_date: z.string(),
+    created_by: z.string().optional(),
+  })
+  .passthrough();
```
kfirstri: use looseObject, passthough is obsolete in zod4

src/cli/commands/entities/records/create.ts
```diff
@@ -0,0 +1,96 @@
...
+                'Example: --data \'{"name": "John", "email": "john@example.com"}\'',
+            },
+          ],
+        },
+      );
+    }
+  }
+
+  if (options.file) {
+    const content = await readTextFile(options.file);
```
kfirstri: we have a `readJsonFile` you can also use

src/cli/commands/entities/records/create.ts
```diff
@@ -0,0 +1,96 @@
...
+
+interface CreateRecordCommandOptions {
+  data?: string;
+  file?: string;
+}
+
+async function parseRecordData(
+  options: CreateRecordCommandOptions,
+): Promise<Record<string, unknown>> {
+  if (options.data) {
```
kfirstri: maybe we can add a hook to the commander to check wether --data or --file but not both? currently if a user passes both just the --data will actually be taken into consideration

src/cli/commands/entities/records/get.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  process.stdout.write(`${JSON.stringify(record, null, 2)}\n`);
+
+  return {};
+}
+
+export function getRecordsGetCommand(context: CLIContext): Command {
+  return new Command("get")
+    .description("Get a single entity record by ID")
+    .argument("<entity-name>", "Name of the entity (e.g. Users, Products)")
```
kfirstri, line 31: Do we want any validation for this thing? I guess the backend will throw if entity does not exists

src/cli/commands/entities/records/update.ts
```diff
@@ -0,0 +1,104 @@
...
+import { InvalidInputError } from "@/core/errors.js";
+import { updateRecord } from "@/core/resources/entity/index.js";
+import { readTextFile } from "@/core/utils/fs.js";
+
+interface UpdateRecordCommandOptions {
+  data?: string;
+  file?: string;
+}
+
+async function parseUpdateData(
```
kfirstri: I see we use this both here and in create.ts , maybe it's worth extracting? (notice we also need to use the readJsonFile method)

src/cli/commands/entities/records/create.ts
```diff
@@ -0,0 +1,96 @@
...
+      return await createRecord(entityName, data);
+    },
+    {
+      successMessage: `Created ${entityName} record`,
+      errorMessage: `Failed to create ${entityName} record`,
+    },
+  );
+
+  log.success(`Record created with ID: ${record.id}`);
+  process.stdout.write(`${JSON.stringify(record, null, 2)}\n`);
```
kfirstri: Let's use normal logs and stuff

you should do 
```
return { outroMessage: `Record created with ID: ${record.id}`}
```

to make it end nicely.
and also lets just log the JSON with the log.success or log.info or soemthing.

This is relevant for all the new commands

src/cli/commands/entities/records/list.ts
```diff
@@ -0,0 +1,59 @@
...
+  process.stdout.write(`${JSON.stringify(records, null, 2)}\n`);
+
+  return {};
+}
+
+export function getRecordsListCommand(context: CLIContext): Command {
+  return new Command("list")
+    .description("List entity records")
+    .argument("<entity-name>", "Name of the entity (e.g. Users, Products)")
+    .option("-f, --filter <json>", "JSON query filter")
```
kfirstri: maybe we need to give a bit mor information here. we also need this to be in the base44 cli SKILL so it will have examples, but lets just say what is expected?

kfirstri (changes requested):
Added some comments :) @claude start working on them please.

src/cli/commands/entities/records/parseRecordData.ts
```diff
@@ -0,0 +1,54 @@
...
+import { readJsonFile } from "@/core/utils/fs.js";
+
+interface RecordDataOptions {
+  data?: string;
+  file?: string;
+}
+
+export async function parseRecordData(
+  options: RecordDataOptions,
+  exampleHint: string,
```
kfirstri: @claude lets make this optional and only use in the exception if available

PR #279, opened 2026-02-19
CLI local dev - entities

artemdemo:
> [!NOTE]
> ## Description
>
> This PR adds a local entity data layer to the dev server, enabling full CRUD operations on app entities during local development without hitting the remote Base44 backend. It introduces an in-memory NeDB database that is seeded from entity definitions, a complete REST API for entity access, and a Socket.IO server that broadcasts real-time events to connected clients on every mutation.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **`database.ts`** (new): `Database` class that creates an in-memory NeDB `Datastore` collection per entity definition
> - **`realtime.ts`** (new): Socket.IO server mounted at `/ws-user-apps/socket.io/` with room-based join/leave, and a `broadcastEntityEvent` helper that emits `update_model` events to `entities:{appId}:{entityName}` rooms
> - **`routes/entities.ts`** (new): Full REST router for entities — `GET /:entityName`, `GET /:entityName/:id`, `POST /:entityName`, `POST /:entityName/bulk`, `PUT /:entityName/:id`, `DELETE /:entityName/:id`, `DELETE /:entityName` — with support for filtering (`q`), sorting, pagination (`limit`/`skip`), and field projection; `GET /User/:id` is forwarded to the remote proxy to preserve auth
> - **`dev-server/main.ts`**: Wires up `Database`, entity routes, and Socket.IO on server start; defers broadcast binding until the HTTP server is ready; closes Socket.IO on shutdown
> - **`commands/dev.ts`**: Loads entities in parallel with functions and passes them to `createDevServer`
> - **`package.json` / `bun.lock`**: Added `@seald-io/nedb` and `socket.io` as dev dependencies
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> All entity IDs are assigned via `nanoid()` on creation. `created_date` and `updated_date` are managed server-side and protected from client overwrite on updates. Internal NeDB `_id` fields are stripped from all responses. The `User` entity is intentionally exempted from local handling and proxied to the remote server to preserve authentication semantics.
>
> ---
> <sub>🤖 Generated by Claude | 2026-02-24 00:00 UTC</sub>

### DIFF
--- a/src/cli/dev/dev-server/routes/entities.ts
+++ b/src/cli/dev/dev-server/routes/entities.ts
@@ -1,3 +1,4 @@
+import type Datastore from "@seald-io/nedb";
 import type { Request, RequestHandler, Response, Router } from "express";
 import { Router as createRouter, json } from "express";
 import { nanoid } from "nanoid";
@@ -75,6 +76,25 @@ export function createEntityRoutes(
   const router = createRouter({ mergeParams: true });
   const parseBody = json();
 
+  function withCollection(
+    handler: (
+      req: Request<EntityParams>,
+      res: Response,
+      collection: Datastore,
+    ) => Promise<void> | void,
+  ): (req: Request<EntityParams>, res: Response) => Promise<void> {
+    return async (req, res) => {
+      const collection = db.getCollection(req.params.entityName);
+      if (!collection) {
+        res
+          .status(404)
+          .json({ error: `Entity "${req.params.entityName}" not found` });
+        return;
+      }
+      await handler(req, res, collection);
+    };
+  }
+
   function emit(
     appId: string,
     entityName: string,
@@ -105,14 +125,8 @@ export function createEntityRoutes(
 
   router.get(
     "/:entityName/:id",
-    async (req: Request<EntityParams>, res: Response) => {
+    withCollection(async (req, res, collection) => {
       const { entityName, id } = req.params;
-      const collection = db.getCollection(entityName);
-
-      if (!collection) {
-        res.status(404).json({ error: `Entity "${entityName}" not found` });
-        return;
-      }
 
       try {
         const doc = await collection.findOneAsync({ id });
@@ -125,19 +139,13 @@ export function createEntityRoutes(
         logger.error(`Error in GET /${entityName}/${id}:`, error);
         res.status(500).json({ error: "Internal server error" });
       }
-    },
+    }),
   );
 
   router.get(
     "/:entityName",
-    async (req: Request<EntityParams>, res: Response) => {
+    withCollection(async (req, res, collection) => {
       const { entityName } = req.params;
-      const collection = db.getCollection(entityName);
-
-      if (!collection) {
-        res.status(404).json({ error: `Entity "${entityName}" not found` });
-        return;
-      }
 
       try {
         const { sort, limit, skip, fields, q } = req.query;
@@ -184,20 +192,14 @@ export function createEntityRoutes(
         logger.error(`Error in GET /${entityName}:`, error);
         res.status(500).json({ error: "Internal server error" });
       }
-    },
+    }),
   );
 
   router.post(
     "/:entityName",
     parseBody,
-    async (req: Request<EntityParams>, res: Response) => {
+    withCollection(async (req, res, collection) => {
       const { appId, entityName } = req.params;
-      const collection = db.getCollection(entityName);
-
-      if (!collection) {
-        res.status(404).json({ error: `Entity "${entityName}" not found` });
-        return;
-      }
 
       try {
         const now = new Date().toISOString();
@@ -218,20 +220,14 @@ export function createEntityRoutes(
         logger.error(`Error in POST /${entityName}:`, error);
         res.status(500).json({ error: "Internal server error" });
       }
-    },
+    }),
   );
 
   router.post(
     "/:entityName/bulk",
     parseBody,
-    async (req: Request<EntityParams>, res: Response) => {
+    withCollection(async (req, res, collection) => {
       const { appId, entityName } = req.params;
-      const collection = db.getCollection(entityName);
-
-      if (!collection) {
-        res.status(404).json({ error: `Entity "${entityName}" not found` });
-        return;
-      }
 
       if (!Array.isArray(req.body)) {
         res.status(400).json({ error: "Request body must be an array" });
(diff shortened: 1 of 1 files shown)


src/cli/dev/dev-server/routes/entities.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+  return Object.keys(projection).length > 0 ? projection : undefined;
+}
+
+function stripInternalFields<T extends Record<string, unknown>>(
+  doc: T,
+): Omit<T, "_id"> {
+  const { _id, ...rest } = doc;
+  return rest;
+}
```
artemdemo, line 68: `_id` is an internal field and is not sent in production

src/cli/dev/dev-server/routes/entities.ts
```diff
@@ -0,0 +1,353 @@
...
+): Record<string, 1 | -1> | undefined {
+  if (!sort) {
+    return undefined;
+  }
+
+  if (sort.startsWith("-")) {
+    return { [sort.slice(1)]: -1 };
+  }
+  return { [sort]: 1 };
+}
```
artemdemo, line 31: According to docs, sorting is a string value, so I'm supporting sorting only by one field.
Also I'm not sure what sorting by multiple fields even mean.

https://docs.base44.com/developers/references/sdk/docs/type-aliases/entities#filter

src/cli/dev/dev-server/routes/entities.ts
```diff
@@ -0,0 +1,354 @@
...
+    async (req: Request<EntityParams>, res: Response) => {
+      const { appId, entityName, id } = req.params;
+      const collection = db.getCollection(entityName);
+
+      if (!collection) {
+        res.status(404).json({ error: `Entity "${entityName}" not found` });
+        return;
+      }
+
+      const { id: _id, created_date, ...body } = req.body;
```
artemdemo: User should not be able to override `id` and `create_date` by sending it in body. So I removing it from the payload.

src/cli/dev/dev-server/routes/entities.ts
```diff
@@ -0,0 +1,365 @@
...
+      } catch (error) {
+        logger.error(`Error in DELETE /${entityName}/${id}:`, error);
+        res.status(500).json({ error: "Internal server error" });
+      }
+    },
+  );
+
+  router.delete(
+    "/:entityName",
+    parseBody,
```
artemdemo, line 322: Apper supports `body` in DELETE method 🤷‍♂️

src/cli/commands/dev.ts
```diff
@@ -18,10 +19,11 @@ async function devAction(options: DevOptions): Promise<RunCommandResult> {
...
       const { project } = await readProjectConfig();
       const configDir = dirname(project.configPath);
-      const functions = await functionResource.readAll(
-        join(configDir, project.functionsDir),
-      );
-      return { functions };
+      const [functions, entities] = await Promise.all([
+        functionResource.readAll(join(configDir, project.functionsDir)),
+        entityResource.readAll(join(configDir, project.entitiesDir)),
+      ]);
```
kfirstri, line 25: just thought about it  - this line `const { project } = await readProjectConfig();`

actually returns all the other resources too, so technitcally this can be - 
```ts
loadResources: async () => {
    const { function, entities } = await readProjectConfig();
    return { functions, entities };
}
```

src/cli/dev/dev-server/main.ts
```diff
@@ -68,6 +79,22 @@ export async function createDevServer(
...
+  // Socket.IO is attached after the HTTP server starts; entity routes receive
+  // a broadcast callback that becomes a no-op until the server is ready.
+  let emitEntityEvent: BroadcastEntityEvent = () => {};
+  const entityRoutes = createEntityRoutes(
+    db,
+    devLogger,
+    remoteProxy,
+    (...args) => emitEntityEvent(...args),
+  );
+  app.use("/api/apps/:appId/entities", entityRoutes);
```
kfirstri, line 96: Should this behave like we did in functions?
only if `if (db.getCollectionNames().length > 0) {` then `createEntityRoutes`

src/cli/dev/dev-server/routes/entities.ts
kfirstri: Seems right to me..

src/cli/dev/dev-server/routes/entities.ts
```diff
@@ -0,0 +1,361 @@
...
+  router.get(
+    "/:entityName/:id",
+    async (req: Request<EntityParams>, res: Response) => {
+      const { entityName, id } = req.params;
+      const collection = db.getCollection(entityName);
+
+      if (!collection) {
+        res.status(404).json({ error: `Entity "${entityName}" not found` });
+        return;
+      }
```
kfirstri: I saw that express has a concept called `router.param` where it's like adding a middleware (callback?) that handles params in url, so i guess we can do something like:

```ts
router.param("entityName", (req, res, next, entityName) => {
    const collection = db.getCollection(entityName);
    if (!collection) {
      res.status(404).json({ error: `Entity "${entityName}" not found` });
      return;
    }
    res.locals.collection = collection;
    next();
  });
```

and then do `const { collection } = res.locals;` in each action.

pros: it encapsulates the collection resolving to one place
cons: Not sure i like passing stuff over res.locals?

wdyt?

src/cli/dev/dev-server/routes/entities.ts
```diff
@@ -0,0 +1,361 @@
...
+
+      try {
+        const now = new Date().toISOString();
+        const { _id, ...body } = req.body;
```
kfirstri, line 206: ~~So the javascript-sdk is generating the _id and sending it in the request?~~
nvm, realized you just extract it out

src/cli/commands/dev.ts
artemdemo: ok

src/cli/dev/dev-server/routes/entities.ts
artemdemo: correct, it's just omitting `_id` from the `body'

src/cli/dev/dev-server/main.ts
artemdemo: I thought about it, and it feels a bit pointless to add this guard.
it cost almost nothing to add it and makes code a bit more straightforward, wdyt?

src/cli/dev/dev-server/routes/entities.ts
artemdemo: I don't like magic of `res.locals` either, but on the other hand I see your point.
It's currently not DRY enough so we can improve by introducing high-order handler - `withCollection`.
Implementing, let me know what do you think.

src/cli/dev/dev-server/main.ts
kfirstri: I agree i think it's fine

PR #321, opened 2026-02-23
feat(secrets): add secrets list, set, and delete commands

Paveltarno:
> [!NOTE]
> ## Description
>
> Adds a `secrets` command group with three subcommands (`list`, `set`, `delete`) for managing project secrets (environment variables) stored in Base44. The `set` command accepts inline `KEY=VALUE` pairs or a `--env-file` path, using the `dotenv` library for parsing. All API responses are validated with Zod schemas following the existing patterns.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `src/core/resources/secret/` module with Zod schemas and API calls for `GET`, `POST`, and `DELETE /api/apps/{appId}/secrets`
> - Added `src/core/utils/env.ts` with `parseEnvFile()` helper using the `dotenv` package
> - Added `src/cli/commands/secrets/` with `list`, `set`, and `delete` subcommands registered under a `secrets` parent command
> - `secrets set` supports both inline `KEY=VALUE` args and `--env-file <path>`; mutual exclusivity is enforced via a Commander `.hook("preAction", ...)` validator
> - Registered `getSecretsCommand` in `src/cli/program.ts`
> - Added `dotenv` to `devDependencies` (bundled at build time)
> - Extended `Base44APIMock` in the testkit with `mockSecretsList`, `mockSecretsSet`, `mockSecretsDelete` (and their error variants)
> - Added 199-line integration test suite (`tests/cli/secrets.spec.ts`) covering happy paths, error paths, and input validation
> - Documented the Commander `.hook("preAction")` validation pattern in `docs/commands.md` and added a copy consistency rule
> - Updated `README.md` command reference table with the three new commands
> - Fixed `.npmrc` registry URL format (added trailing slash)
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> Values with embedded `=` signs (e.g. connection strings) are handled correctly because `parseEntries` splits only on the **first** `=` character via `indexOf`. The `--env-file` path is resolved to an absolute path before parsing so relative paths work regardless of the current working directory at call time.
>
> ---
> <sub>🤖 Generated by Claude | 2026-02-24 15:45 UTC</sub>


src/core/resources/secret/api.ts
```diff
@@ -0,0 +1,87 @@
...
+  const appClient = getAppClient();
+
+  let response: KyResponse;
+  try {
+    response = await appClient.get("secrets");
+  } catch (error) {
+    throw await ApiError.fromHttpError(error, "listing secrets");
+  }
+
+  const result = ListSecretsResponseSchema.safeParse(await response.json());
```
kfirstri, line 25: i see that the `ListSecretsResponse` is a `z.record(z.string(), z.string())`, 
then when we do the "secrets list" command we only grab the keys.. but the values here are just censored values.. maybe this `listSecret` function can just return an array of string instead of an object?

src/cli/commands/secrets/list.ts
```diff
@@ -0,0 +1,41 @@
...
+    async () => {
+      return await listSecrets();
+    },
+    {
+      successMessage: "Secrets fetched successfully",
+      errorMessage: "Failed to fetch secrets",
+    },
+  );
+
+  const names = Object.keys(secrets);
```
kfirstri, line 20: also wrote it in the api.ts file - can the `listSecrets` function just return an array of names instead of an object?

src/cli/commands/secrets/list.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  if (names.length === 0) {
+    return { outroMessage: "No secrets configured." };
+  }
+
+  for (const [name, maskedValue] of Object.entries(secrets)) {
+    log.info(`${name} = ${maskedValue}`);
```
kfirstri: i don't see the value of showing masked values..? wdyt

src/cli/commands/secrets/set.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+async function setSecretsAction(
+  entries: string[],
+  options: { envFile?: string },
+): Promise<RunCommandResult> {
+  const hasEntries = entries.length > 0;
+  const hasEnvFile = Boolean(options.envFile);
+
+  if (!hasEntries && !hasEnvFile) {
```
kfirstri, line 43: move these validations outside the setSecretAction, you can use a commander hook. maybe also worth documenting this

src/core/resources/secret/api.ts
Paveltarno: we also grab the values (this is per the design), these are just `xxxx` like sequences but they do have the original length of the value

src/cli/commands/secrets/set.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (hasEntries && hasEnvFile) {
+    throw new InvalidInputError(
+      "Provide KEY=VALUE pairs or --env-file, but not both.",
+    );
+  }
+
+  let secrets: Record<string, string>;
+
+  if (hasEnvFile) {
+    secrets = await parseEnvFile(options.envFile as string);
```
kfirstri: does this work relative? do we need to resolve the path we got from the user? lets test this

src/cli/commands/secrets/set.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const names = Object.keys(secrets);
+
+  await runTask(
+    `Setting ${names.length} secret${names.length === 1 ? "" : "s"}`,
+    async () => {
+      return await setSecrets(secrets);
+    },
+    {
+      successMessage: `${names.length} secret${names.length === 1 ? "" : "s"} set successfully`,
```
kfirstri: this is the 3rd time i see this names.length=== 1 check.. lets think about removing it? extract it to a function?

src/cli/commands/secrets/set.ts
```diff
@@ -0,0 +1,101 @@
...
+    const value = entry.slice(eqIndex + 1);
+
+    if (!key) {
+      throw new InvalidInputError(
+        `Invalid format: "${entry}". Key cannot be empty.`,
+      );
+    }
+
+    secrets[key] = value;
+  }
```
kfirstri, line 32: Not sure where, i see in the dashboard there's some validation

Image

Maybe we can enforce it in the API layer with zod / here with some validations?

src/core/utils/env.ts
```diff
@@ -0,0 +1,12 @@
+import { readFile } from "node:fs/promises";
```
kfirstri: use the fs.ts readfile function

src/cli/commands/secrets/list.ts
Paveltarno: yes

src/cli/commands/secrets/set.ts
Paveltarno: Decided not to add a pluralize helper — instead dropped the singular/plural logic entirely and always use the plural form, which aligns with the copy style of other commands.

src/cli/commands/secrets/set.ts
Paveltarno: very fair point, i actually looked into this a bit. It might actually put us in a bit of a weird spot as the LLM or other parts of `apper` CAN create secrets in any format. I'm afaid that actually having this validation can potentially "lock the user". Imagine this sceanrio:
1. LLM creates a "my spaced secret" key, on `apper`.
2. user ejects, starts using cli and wants to update the secret
3. they can't because we don't let them, they can delete it but some other parts of their systems might need rewriting now

src/cli/commands/secrets/set.ts
Paveltarno: good idea!

src/cli/commands/secrets/set.ts
Paveltarno: good catch! 🐟

src/core/utils/env.ts
Paveltarno: Done — switched to `readTextFile` from `fs.ts`. Also resolved the relative path at the CLI boundary (`set.ts`) rather than in the core layer, so `parseEnvFile` always receives an absolute path like all other places in the system

docs/commands.md
```diff
@@ -167,10 +167,39 @@ theme.format.agentHints(hints)            // "[Agent Hints]\n  Run: ..."
...
+    .option("--flag-a <value>", "Alternative input")
+    .hook("preAction", validateInput)
+    .action(async (entries, options) => {
+      await runCommand(() => myAction(entries, options), { requireAuth: true }, context);
+    });
+}
+```
+
+Access `command.args` for positional arguments and `command.opts()` for options inside the hook. See `secrets/set.ts` and `project/create.ts` for real examples.
+
```
kfirstri, line 197: love this

.npmrc
```diff
@@ -1 +1 @@
-npm_config_registry=https://registry.npmjs.org
\ No newline at end of file
+registry=https://registry.npmjs.org/
```
kfirstri, line 1: why this change?

Paveltarno: ~because my default repo is set to wix's, I remember you talked about this and didn't realize this never happened~

I now realize it happened (didn't see the file locally for some reason)
so this is an interesting glitch where if bun sees `npm_config_VALUE` it will try and set it as an env var, but it won't if this env var is already set in the `printenv` array. It WILL however override that if you write down the specific value (like `registry`)
this works well with the `npmpublic/npmprivate` aliasies we have set up at wix

src/cli/commands/secrets/set.ts
kfirstri: lets say looooose

PR #353, opened 2026-03-01
Watching and reloading functions

artemdemo:
> [!NOTE]
> ## Description
>
> This PR adds hot-reload support for backend functions in the local dev server. When function files change on disk, the dev server now automatically detects the change and reloads the function definitions without requiring a manual restart. A new `WatchBase44` class backed by `chokidar` handles filesystem watching with debouncing, graceful recovery when watched directories are deleted and recreated, and proper cleanup on shutdown.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `src/cli/dev/dev-server/watcher.ts` — new `WatchBase44` class using `chokidar` v5 to watch filesystem paths, with debounced change callbacks, a creation-queue for paths that don't yet exist, and watcher garbage collection on directory deletion
> - Added `FunctionManager.reload()` method to atomically stop all running functions and replace the function map with a fresh set
> - Fixed zombie process bug in `FunctionManager`: if `stopAll()` is called while a function process is still initializing, the process is now properly killed after the start promise resolves
> - Improved `FunctionManager` exit logging: signal-terminated processes (code === null) no longer log a spurious exit message
> - Wired up `WatchBase44` in `main.ts` to watch the configured functions directory and call `functionManager.reload()` on any change
> - Moved function route registration outside the `if`-block so routes are always registered regardless of initial function count
> - Refactored `createDevServer` promise chain to `await` the server start and flatten the remaining setup code
> - Replaced `lodash.kebabcase` (single-function package) with `lodash` and updated imports to `lodash/kebabCase`
> - Added `chokidar` and `@types/lodash` as dev dependencies; removed `lodash.kebabcase` and `@types/lodash.kebabcase`
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The watcher uses a 300 ms debounce to coalesce rapid file-system events (e.g. editor save storms) into a single reload. Directories that don't exist at startup are polled every 500 ms until they appear, enabling the watcher to survive a full functions-directory removal and recreation at runtime.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-04 09:30 UTC</sub>

### DIFF
--- a/src/cli/dev/dev-server/main.ts
+++ b/src/cli/dev/dev-server/main.ts
@@ -161,17 +161,23 @@ export async function createDevServer(
     devLogger,
   );
   base44ConfigWatcher.on("change", async (name) => {
-    if (name === "functions") {
-      const { functions } = await options.loadResources();
-      const previousFunctionCount = functionManager.getFunctionNames().length;
-      functionManager.reload(functions);
+    try {
+      if (name === "functions") {
+        const { functions } = await options.loadResources();
+        const previousFunctionCount = functionManager.getFunctionNames().length;
+        functionManager.reload(functions);
 
-      const names = functionManager.getFunctionNames();
-      if (names.length > 0) {
-        devLogger.log(`Reloaded functions: ${names.sort().join(", ")}`);
-      } else if (previousFunctionCount > 0) {
-        devLogger.log("All functions removed");
+        const names = functionManager.getFunctionNames();
+        if (names.length > 0) {
+          devLogger.log(`Reloaded functions: ${names.sort().join(", ")}`);
+        } else if (previousFunctionCount > 0) {
+          devLogger.log("All functions removed");
+        }
       }
+    } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+      devLogger.error(errorMessage);
     }
   });
   await base44ConfigWatcher.start();


src/cli/commands/project/create.ts
```diff
@@ -3,7 +3,7 @@ import type { Option } from "@clack/prompts";
 import { confirm, group, isCancel, log, select, text } from "@clack/prompts";
 import { Argument, Command } from "commander";
 import { execa } from "execa";
-import kebabCase from "lodash.kebabcase";
+import kebabCase from "lodash/kebabCase";
```
artemdemo, line 6: I installed the whole `lodash`, so in order to provide explicit tree-shacking - changing import.

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -98,6 +98,11 @@ export class FunctionManager {
     return this.waitForReady(name, runningFunc);
   }
 
+  reload(functions: BackendFunction[]): void {
+    this.stopAll();
+    this.functions = new Map(functions.map((f) => [f.name, f]));
+  }
```
artemdemo, line 110: Basically reload is turning down all functions and replacing list of available functions.
Then when request will hit dev server - required function will be instantiated in Deno.
(I'm not doing HMR, which imo pointless in this case and not sure how feasible)

src/cli/dev/dev-server/main.ts
```diff
@@ -23,6 +27,30 @@ import {
 
 const DEFAULT_PORT = 4400;
 const BASE44_APP_URL = "https://base44.app";
+const WATCH_DEBOUNCE_MS = 300;
+
+async function watchDir(
```
artemdemo: I'm planning to use this method for other watchers (like entities), so making it relatively generic.

src/cli/dev/dev-server/main.ts
```diff
@@ -23,6 +27,30 @@ import {
...
+      await onChange();
+    } catch (error) {
+      logger.error(
+        `[dev-server] Watch handler failed for ${dir}`,
+        error instanceof Error ? error : undefined,
+      );
+    }
+  }, WATCH_DEBOUNCE_MS);
+
+  return watch(dir, { recursive: true }, debouncedOnChange);
```
artemdemo: Watching for everything in the folder. I don't care about specific files, this folder should have either logic (`ts` files) or configuration (`json` files) - change in either of them should trigger reload.

src/cli/dev/dev-server/main.ts
```diff
@@ -75,14 +103,13 @@ export async function createDevServer(
   const devLogger = createDevLogger();
 
   const functionManager = new FunctionManager(functions, devLogger);
+  const functionRoutes = createFunctionRouter(functionManager, devLogger);
+  app.use("/api/apps/:appId/functions", functionRoutes);
```
artemdemo, line 81: Moved because functions now can be added into the project that didn't have them previously.

artemdemo:
@claude  review

src/cli/dev/dev-server/watcher.ts
```diff
@@ -0,0 +1,96 @@
+import { basename, normalize, relative } from "node:path";
+import { type FSWatcher, watch } from "chokidar";
```
artemdemo, line 3: At first I wanted to use `watch` from node, but turns out on Linux it will not watch recursively the folder.
This feature is available fully (with all bug fixes and features) from Node 22 (which is higher than we currently support).
`chokidar` is cross platform by design.

src/cli/dev/dev-server/watcher.ts
```diff
@@ -0,0 +1,116 @@
+import { relative } from "node:path";
+import { type FSWatcher, watch } from "chokidar";
+import debounce from "lodash/debounce";
+import { pathExists } from "@/core/utils/fs.js";
+import type { Logger } from "../createDevLogger.js";
+
+const WATCH_DEBOUNCE_MS = 300;
+const WATCH_QUEUE_DELAY_MS = 500;
+
+type WatchEntryName = "functions";
```
artemdemo, line 11: I will add `entries` here in the next PR

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -75,6 +75,12 @@ export class FunctionManager {
     try {
       return await promise;
     } finally {
+      if (!this.starting.has(name) && this.running.has(name)) {
+        // We can end up here if user called `stopAll()` while one of the processes is being initialized.
+        // In this case we need to kill the process, to avoid zombies running around.
+        this.running.get(name)?.process.kill();
```
kfirstri, line 81: Should we grab the returned boolean and only if true delete it from the running set? I guess we can get `false` from the `.kill()` and still do `this.running.delete(name)` which might place us in a weird state?

src/cli/dev/dev-server/function-manager.ts
```diff
@@ -148,9 +159,14 @@ export class FunctionManager {
...
-        `[dev-server] Function "${name}" exited with code ${code}`,
-      );
+      // `code === null` happens when process is terminated by a signal.
+      // In this case I'm assuming that it's happening as part of the reload mechanism.
+      // In other words there is no need to log `code` information, if we ourselve killed the process.
+      if (code !== null) {
+        this.logger.log(
+          `[dev-server] Function "${name}" exited with code ${code}`,
+        );
+      }
```
kfirstri, line 165: thought: maybe we can indicate this with `this.logger.debug` that the function died becuase of a reload?
Not sure it's interesting, i see that you already print in main.ts 
```
devLogger.log(`Reloaded functions: ${names.sort().join(", ")}`);
```

src/cli/dev/dev-server/watcher.ts
kfirstri: I guess you mean entities

artemdemo: yes

src/cli/dev/dev-server/function-manager.ts
artemdemo: I indicate that functions have been changed as a result of reload.
This mainly because when I see it as developer - I feel better, it's proof that process didn't die :)
IMO it's good enough for now. Let me know if you think otherwise.

src/cli/dev/dev-server/function-manager.ts
artemdemo: The whole scenario that this `if` represents is unlikely. It can happen in systems with the heigh load, but locally it's very rare.
Therefore I suggest to pass on handling additional rare scenario of returned `false` :)

src/cli/dev/dev-server/watcher.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  async start(): Promise<void> {
+    if (this.watchers.length > 0 || this.queueWaitForCreation.length > 0) {
+      return;
+    }
+    for (const item of this.itemsToWatch) {
+      if (await pathExists(item.path)) {
+        this.watchers.push(this.watchTarget(item));
+      } else {
+        this.queueWaitForCreation.push(item);
```
kfirstri, line 42: I want to be sure i understand - this is implemented so we don't have to create watchers on non-existing folders? Technically we can just create the `this.watchers` on the non-existing path and it will "emit" once created, no?

src/cli/dev/dev-server/watcher.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  private watchTarget(item: WatchEntry): FSWatcher {
+    const handler = debounce(async (_event: string, path: string) => {
+      try {
+        await this.onChange(item.name, relative(item.path, path));
+      } catch (err) {
+        this.logger.error(
+          `Reload failed for ${item.name}`,
+          err instanceof Error ? err : undefined,
+        );
```
kfirstri: I feel that this `await this.onChange` and error catching couples the Watcher with whatever its "listeners" are doing.. watcher only alerts that there was a change, not sure it should log error "reload failed" if onchange failed.

Some thought - maybe keep the same vibe like chokidar, have `WatchBase44` inherit the `EventEmitter` class, and just `this.emit('change', item.name, relativePath)`, this way in main.ts after creating `base44ConfigWatcher` you can do `base44ConfigWatcher.on('change'....` and reload and everything else.

wdyt? too complicated?

src/cli/dev/dev-server/watcher.ts
```diff
@@ -0,0 +1,116 @@
...
+        // Garbage collecting closed watchers.
+        // This needs to happen in the next "tick", so process will not be "stuck".
+        // (User wouldn't be able to exit otherwise)
+        this.watchers = this.watchers.filter((watcher) => !watcher.closed);
+      });
+    });
+    watcher.on("error", (err) => {
+      this.logger.error(
+        `[dev-server] Watch handler failed for ${item.path}`,
+        err instanceof Error ? err : undefined,
```
kfirstri, line 110: what happens other then logging? should we recreate the watcher or something?

src/cli/dev/dev-server/main.ts
```diff
@@ -140,24 +141,49 @@ export async function createDevServer(
...
+      if (name === "functions") {
+        const { functions } = await options.loadResources();
+        const previousFunctionCount = functionManager.getFunctionNames().length;
+        functionManager.reload(functions);
+
+        const names = functionManager.getFunctionNames();
+        if (names.length > 0) {
+          devLogger.log(`Reloaded functions: ${names.sort().join(", ")}`);
+        } else if (previousFunctionCount > 0) {
+          devLogger.log("All functions removed");
```
kfirstri, line 174: General comment about logs - I noticed some weird behavior with the prefixes - 
* Dev server logs - here we don't have a `[dev-server]` prefix, but sometimes we do have it
* Function logs - the deno-runtime has `console.log(`[${functionName}] ` but the output it writes is logged without a prefix

src/cli/dev/dev-server/watcher.ts
artemdemo: tbh, not sure how this case is recoverable.
I suspect that it can fail mainly because of file permissions and in this case there is nothing we can recover.
So maybe we just need to inform the user and maximum see if there will be complains

src/cli/dev/dev-server/watcher.ts
artemdemo: no, watcher will not be created for non existing path, we need to wait up until the path is actually created.

src/cli/dev/dev-server/watcher.ts
artemdemo: no, let's do it.

src/cli/dev/dev-server/main.ts
artemdemo: tbh, I'm not sure `[dev-server]` prefix makes sense - everything is basically "dev-server".
Removing it.

PR #357, opened 2026-03-02
feat: add `base44 exec` command

netanelgilad:
> [!NOTE]
> ## Description
>
> Adds a new `base44 exec` command that allows users to run TypeScript/JavaScript scripts with the Base44 SDK pre-authenticated and available as a global `base44` variable. Scripts are piped via stdin and executed using Deno as the runtime, with the CLI handling token exchange and SDK setup automatically.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Add `base44 exec` CLI command (`packages/cli/src/cli/commands/exec.ts`) that reads a script from stdin and runs it via Deno
> - Add `runScript` core function (`packages/cli/src/core/exec/run-script.ts`) that spawns Deno with the exec wrapper, passing auth tokens via environment variables
> - Add Deno exec wrapper (`packages/cli/deno-runtime/exec.ts`) that creates a pre-authenticated `@base44/sdk` client and exposes it as a global `base44` before importing the user's script
> - Add `getAppUserToken()` and `getSiteUrl()` to `packages/cli/src/core/project/api.ts` for token exchange and site URL resolution
> - Add `verifyDenoInstalled()` utility (`packages/cli/src/core/utils/dependencies.ts`) that gives a user-friendly error with install instructions if Deno is missing
> - Register exec command in `program.ts` and include the Deno runtime asset in the build pipeline
> - Add Deno to the CI workflow (`test.yml`) so integration tests can actually spawn Deno
> - Add comprehensive integration tests (`tests/cli/exec.spec.ts`) covering stdin validation, token exchange errors, SDK availability, auth headers, and exit code forwarding
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The exec wrapper is copied to a temporary location outside `node_modules` before being passed to Deno. This is required for Deno 2.x compatibility — files inside `node_modules` are treated as Node modules and block `npm:` specifiers. The wrapper uses `--node-modules-dir=auto` so Deno can resolve `npm:@base44/sdk` without a lock file.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-18 00:00 UTC</sub>

### DIFF
--- a/packages/cli/src/cli/commands/exec.ts
+++ b/packages/cli/src/cli/commands/exec.ts
@@ -57,6 +57,16 @@ export function getExecCommand(): Command {
     .description(
       "Run a script with the Base44 SDK pre-authenticated as the current user",
     )
+    .addHelpText(
+      "after",
+      `
+Examples:
+  Run a script file:
+    $ cat ./script.ts | base44 exec
+
+  Inline script:
+    $ echo "const users = await base44.entities.User.list()" | base44 exec`,
+    )
     .action(async (_options: unknown, command: Base44Command) => {
       return await execAction(command.isNonInteractive);
     });


infra/build.ts
```diff
@@ -34,9 +34,15 @@ const runAllBuilds = async () => {
     entrypoints: ["./deno-runtime/main.ts"],
     outdir: "./dist/deno-runtime",
   });
+  const denoExec = await runBuild({
```
netanelgilad: Do we really need to build the deno-runtime exec ts? we can use ts directly in deno nope?

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,199 @@
...
+function verifyDenoIsInstalled(): void {
+  const result = spawnSync("deno", ["--version"]);
+  if (result.error) {
+    throw new DependencyNotFoundError(
+      "Deno is required to run scripts with exec",
+      {
+        hints: [
+          {
+            message:
+              "Install Deno: https://docs.deno.com/runtime/getting-started/installation/",
```
netanelgilad:
```suggestion
              "Install Deno: https://docs.deno.com/runtime/getting_started/installation/",
```

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,199 @@
...
+  return {};
+}
+
+export function getExecCommand(context: CLIContext): Command {
+  const cmd = new Command("exec")
+    .description(
+      "Run a script with the Base44 SDK pre-authenticated as the current user",
+    )
+    .argument("[script]", "Path to a .ts or .js script file")
+    .option("-e, --eval <code>", "Evaluate inline code")
```
netanelgilad: let's have an explicit flag for stdin instead of inferring it?

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,199 @@
...
+    );
+  }
+
+  // Fetch the app's published URL (subdomain) so the SDK can route
+  // function invocations through the app domain instead of the platform.
+  let appBaseUrl: string | undefined;
+  try {
+    appBaseUrl = await getSiteUrl();
+  } catch {
+    // Non-fatal: fall back to the platform API URL
```
netanelgilad: Nah, it's fatal, let's fail on it. it should make the call to the site url to avoid function invocation issues

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,199 @@
...
+      error,
+      "exchanging platform token for app user token",
+    );
+  }
+
+  // Fetch the app's published URL (subdomain) so the SDK can route
+  // function invocations through the app domain instead of the platform.
+  let appBaseUrl: string | undefined;
+  try {
+    appBaseUrl = await getSiteUrl();
```
netanelgilad: do this and the above token exchange in parallel

deno-runtime/exec.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+import { createClient } from "npm:@base44/sdk";
+
+// Use the app's published URL (subdomain) as serverUrl when available so that
+// function invocations route through the app domain instead of the platform.
+const base44 = createClient({
+  appId,
+  token: accessToken,
+  serverUrl: appBaseUrl || apiUrl || "https://app.base44.com",
```
netanelgilad: let's always use the appBaseUrl, remove other fallbacks, not needed.

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+      }
+    })(),
+    getSiteUrl(),
+  ]);
+
+  // Copy the exec wrapper out of node_modules to a temp location.
+  // Deno 2.x treats files inside node_modules as Node modules and
+  // doesn't support npm: specifiers in them.
+  const tempWrapper = join(tmpdir(), `base44-exec-wrapper-${Date.now()}.ts`);
+  copyFileSync(EXEC_WRAPPER_PATH, tempWrapper);
```
artemdemo: Does it mean we not only require Deno, but also Deno 2 and above?
Should we then check for that?

infra/build.ts
```diff
@@ -34,6 +34,9 @@ const runAllBuilds = async () => {
     entrypoints: ["./deno-runtime/main.ts"],
     outdir: "./dist/deno-runtime",
   });
+  // Deno runs TypeScript natively, so just copy exec.ts as-is (no bundling needed)
+  mkdirSync("./dist/deno-runtime", { recursive: true });
+  copyFileSync("./deno-runtime/exec.ts", "./dist/deno-runtime/exec.ts");
```
artemdemo: Should then both "runtimes" be aligned then?

src/cli/commands/exec.ts
artemdemo: In linux apps, there is inferring, and I don't see apps that have `--stdin` flag.
What problem do you think inferring is introducing?

packages/cli/src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+        }
+      }
+    }
+  }
+
+  return {};
+}
+
+export function getExecCommand(context: CLIContext): Command {
+  const cmd = new Command("exec")
```
artemdemo: why `exec` command has so many options?
Is it grounded in any user voice?
Maybe it's better to make it simpler in the beginning and start with file path and then introduce additional options?
It's easier to add such functionality, then to remove flags that may be in use.

src/cli/commands/exec.ts
artemdemo: I would even argue that if your main user is an agent, then maybe `-e` flag and file path are not needed - it should be only `stdin`.
any system will allow you to read a file, echo any code and pass it as `stdin`

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+        hints: [
+          { message: "File:  base44 exec ./script.ts" },
+          { message: 'Eval:  base44 exec -e "console.log(1)"' },
+          { message: "Stdin: echo 'code' | base44 exec --stdin" },
+        ],
+      },
+    );
+  }
+
+  verifyDenoIsInstalled();
```
kfirstri: Please move all the validation to a commander hook (you can see other commands do it)

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+          .get("auth/token")
+          .json<{ token: string }>();
+        return response.token;
+      } catch (error) {
+        throw await ApiError.fromHttpError(
+          error,
+          "exchanging platform token for app user token",
+        );
+      }
+    })(),
```
kfirstri: Please create a new function to do it inside core/ , I don't want cli to be that familiar with how we create an appClient, just export a function called get user app token or something and call it here

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+  try {
+    const exitCode = await new Promise<number>((resolvePromise) => {
+      const child = spawn(
+        "deno",
+        [
+          "run",
+          "--allow-all",
+          "--node-modules-dir=auto",
+          tempWrapper,
+          ...extraArgs,
```
kfirstri: Do we really want to give the user more controls in the deno instance? Isn't this too coupled to deno

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
+import { spawn, spawnSync } from "node:child_process";
+import { copyFileSync, unlinkSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
```
kfirstri: we use the "tmp-promise" package everywhere needed, so let's use it here too, I think it also give you a cleanup

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+  DependencyNotFoundError,
+  InvalidInputError,
+} from "@/core/errors.js";
+import { getAppConfig } from "@/core/project/app-config.js";
+import { getSiteUrl } from "@/core/site/api.js";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const EXEC_WRAPPER_PATH = join(__dirname, "../deno-runtime/exec.ts");
+
+function verifyDenoIsInstalled(): void {
```
kfirstri: Don't we already have this check for local dev? Maybe it can be reused? Not critical

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+      data += chunk;
+    });
+    process.stdin.on("end", () => resolve(data));
+    process.stdin.on("error", reject);
+  });
+}
+
+interface ExecOptions {
+  eval?: string;
+  stdin?: boolean;
```
kfirstri: I'm not sure if we need this, can't we just pipe something inside?

src/cli/commands/exec.ts
```diff
@@ -0,0 +1,196 @@
...
+          env: {
+            ...process.env,
+            SCRIPT_PATH: scriptPath,
+            BASE44_APP_ID: appConfig.id,
+            BASE44_ACCESS_TOKEN: appUserToken,
+            BASE44_APP_BASE_URL: appBaseUrl,
+          },
+          stdio: "inherit",
+        },
+      );
```
kfirstri: We need to test this one windows

packages/cli/src/cli/program.ts
```diff
@@ -66,6 +67,9 @@ export function createProgram(context: CLIContext): Command {
   // Register types command
   program.addCommand(getTypesCommand(context));
 
+  // Register exec command
+  program.addCommand(getExecCommand(context));
```
kfirstri: Consider releasing this as hidden first since there's no docs or skills

packages/cli/tests/cli/exec.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    t.api.mockError("get", `/api/apps/test-app-id/auth/token`, {
+      status: 500,
+      body: { error: "Internal server error" },
+    });
+
+    const result = await t.run("exec", "-e", "console.log(1)");
+
+    t.expectResult(result).toFail();
+  });
+});
```
kfirstri, line 103: Technically we can add some success tests with mock and such

src/cli/commands/exec.ts
artemdemo: I think it's critical :) we should use same check

infra/build.ts
netanelgilad: hmm you mean the one for backend functions and the one for exec?
maybe we should align them as close as possible, but they are not exactly the same, as one works with Deno.serve (be function) and one is just a script

src/cli/commands/exec.ts
netanelgilad: hmm I agree on stdin probably being enough for all, but there is also the `-` convention for stdin:
https://chatgpt.com/share/69ad1952-accc-800a-a3d9-0729403652db

@claude
what do you think? what would be the best convention here so AI would understand this interface the best?

netanelgilad: actually the prior art is the node cli itself, which accepts `-e` for quick eval, `-` for stdin, and file path as positional arg

packages/cli/src/cli/commands/exec.ts
netanelgilad: the idea was to match the `node` cli experience for prior art and expected behavior from common clis that have the same pattern.
I'm not sure we'll have clear user voice about it so I added the common cli experience 🤔

infra/build.ts
artemdemo: I was referring to copying vs bundling, not the actual code.

packages/cli/src/cli/commands/exec.ts
artemdemo: then I'm in favor to simplifying. Many options is API for humans, but agent doesn't care what to use and how, then imo simplest way with single option is the best.

infra/build.ts
netanelgilad: ah got u sry.
what do you think then about copying? seem preferred nope?
If you think it's good I can make a small unrelated PR to move to just copying

artemdemo: actually yes - copying look simpler solution. If it works, then amazing :)

netanelgilad: aligned in another PR

packages/cli/src/cli/commands/exec.ts
netanelgilad: went only with stdin option for now, works for all cases, piping a file, piping "echo" for eval etc

src/cli/commands/exec.ts
netanelgilad: how do we test? can we use github's matrix tests?

packages/cli/src/core/utils/app-info.ts
kfirstri, line 1: the core/utils folder was more intended to small utility function for the other resources and files, i would rather have api calls like here to be in core/project folder?

packages/cli/src/core/utils/app-info.ts
```diff
@@ -0,0 +1,40 @@
...
+    return response.token;
+  } catch (error) {
+    throw await ApiError.fromHttpError(
+      error,
+      "exchanging platform token for app user token",
+    );
+  }
+}
+
+export async function getSiteUrl(projectId?: string): Promise<string> {
```
kfirstri: can't this stay under core/site/api? or maybe core/project as well?

src/cli/commands/exec.ts
kfirstri: No windows yet.. i just meant to check it manually on windows to see that everything works fine

packages/cli/src/core/exec/run-script.ts
```diff
@@ -0,0 +1,76 @@
...
+  verifyDenoInstalled("to run scripts with exec");
+
+  const cleanupFns: (() => void)[] = [];
+
+  const tempScript = await file({ postfix: ".ts" });
+  cleanupFns.push(tempScript.cleanup);
+  writeFileSync(tempScript.path, code, "utf-8");
+  const scriptPath = `file://${tempScript.path}`;
+
+  const appConfig = getAppConfig();
```
kfirstri: I'm working on some refactoring the `getAppConfig()`, can you call it in your command, and just pass the appId into the runScript function? (it will be provided later by the CLIContext in the command so im trying to minimize dependencies to getAppConfig)

packages/cli/src/cli/commands/exec.ts
```diff
@@ -0,0 +1,55 @@
...
+    "No input provided. Pipe a script to stdin.",
+    {
+      hints: [
+        { message: "File:  cat ./script.ts | base44 exec" },
+        { message: 'Eval:  echo "console.log(1)" | base44 exec' },
+      ],
+    },
+  );
+
+  if (process.stdin.isTTY) {
```
kfirstri: Lets use the command.isNonInteractive, you can see other commands passing it into the action function and using it

packages/cli/src/cli/commands/exec.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+}
+
+async function execAction(): Promise<RunCommandResult> {
+  const noInputError = new InvalidInputError(
+    "No input provided. Pipe a script to stdin.",
+    {
+      hints: [
+        { message: "File:  cat ./script.ts | base44 exec" },
+        { message: 'Eval:  echo "console.log(1)" | base44 exec' },
```
kfirstri: maybe we can give a better example? like `console.log(base44.entities.list())`?

packages/cli/src/cli/commands/exec.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  return {};
+}
+
+export function getExecCommand(): Command {
+  return new Base44Command("exec")
+    .description(
+      "Run a script with the Base44 SDK pre-authenticated as the current user",
+    )
```
kfirstri, line 59: consider adding examples (here's an example from the create command)
```ts
    .addHelpText(
      "after",
      `
Examples:
  $ base44 create my-app                                         Creates a base44 project at ./my-app
  $ base44 create my-todo-app --template backend-and-client      Creates a base44 backend-and-client project at ./my-todo-app
  $ base44 create my-app --path ./projects/my-app --deploy       Creates a base44 project at ./project/my-app and deploys it`,
    )
```

packages/cli/src/cli/commands/exec.ts
```diff
@@ -0,0 +1,55 @@
...
+  const code = await readStdin();
+
+  if (!code.trim()) {
+    throw noInputError;
+  }
+
+  const { exitCode } = await runScript({ code });
+
+  if (exitCode !== 0) {
+    process.exitCode = exitCode;
```
kfirstri, line 49: I'll open an issue about command should be able to return a `exitCode`, instead of messing with process.exitCode.. but it's fine for now

packages/cli/src/cli/commands/exec.ts
```diff
@@ -0,0 +1,55 @@
...
+    throw noInputError;
+  }
+
+  const code = await readStdin();
+
+  if (!code.trim()) {
+    throw noInputError;
+  }
+
+  const { exitCode } = await runScript({ code });
```
kfirstri: do we need any nicer try/catch?

kfirstri: oh the try/catch is inside the wrapper.. so maybe it's fine

PR #364, opened 2026-03-04
Watching Entities

artemdemo:
> [!NOTE]
> ## Description
>
> This PR adds live-reload support for the entities directory in the dev server. Previously, only the functions directory was watched for changes; now the entities directory is also watched, and when it changes, the in-memory database is cleared and reloaded automatically. The watcher was refactored to be generic and handle root-directory deletion correctly.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **`database.ts`**: Converted `Database` from constructor-based initialization to a `load(entities)` method pattern; added `dropAll()` to clear all in-memory collections.
> - **`main.ts`**: Updated `Database` instantiation to use the new `load()` API; added the entities directory to the file watcher; added an `"entities"` change handler that drops and reloads the database when entity files change.
> - **`watcher.ts`**: Made `WatchBase44` generic (`WatchBase44<T extends string>`) so it accepts any named watch targets via a `Record<T, string>` map instead of a typed array; replaced per-target `unlinkDir` root-deletion handling with a unified polling loop (`watchEntries`) that detects both creation and deletion of watched root directories; switched internal watcher storage from an array to a `Map<T, FSWatcher>` for O(1) lookup by name.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The watcher refactor addresses a Chokidar limitation: it does not fire an `unlinkDir` event when the **root** directory being watched is deleted (only for nested dirs). The new `watchEntries` polling loop detects this case and emits a `"change"` event accordingly, then cleans up closed watchers asynchronously to avoid blocking the Node.js event loop.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-04 00:00 UTC</sub>

### DIFF
--- a/src/cli/dev/dev-server/database.ts
+++ b/src/cli/dev/dev-server/database.ts
@@ -2,10 +2,9 @@ import Datastore from "@seald-io/nedb";
 import type { Entity } from "@/core/resources/entity/schema.js";
 
 export class Database {
-  private collections: Map<string, Datastore>;
+  private collections: Map<string, Datastore> = new Map();
 
-  constructor(entities: Entity[]) {
-    this.collections = new Map();
+  load(entities: Entity[]) {
     for (const entity of entities) {
       this.collections.set(entity.name, new Datastore());
     }
@@ -17,5 +16,12 @@ export class Database {
 
   getCollectionNames(): string[] {
     return Array.from(this.collections.keys());
+  }
+
+  dropAll() {
+    for (const collection of this.collections.values()) {
+      collection.remove({}, { multi: true });
+    }
+    this.collections.clear();
   }
 }
--- a/src/cli/dev/dev-server/main.ts
+++ b/src/cli/dev/dev-server/main.ts
@@ -86,7 +86,8 @@ export async function createDevServer(
     );
   }
 
-  const db = new Database(entities);
+  const db = new Database();
+  db.load(entities);
   if (db.getCollectionNames().length > 0) {
     clackLog.info(`Loaded entities: ${db.getCollectionNames().join(", ")}`);
   }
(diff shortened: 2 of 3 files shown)


src/cli/dev/dev-server/main.ts
```diff
@@ -152,12 +153,16 @@ export async function createDevServer(
...
+      functions: {
         name: "functions",
         path: join(dirname(project.configPath), project.functionsDir),
       },
-    ],
+      entities: {
+        name: "entities",
+        path: join(dirname(project.configPath), project.entitiesDir),
+      },
+    },
```
artemdemo, line 159: I want this configuration to be an object so there will be no chance of passing "functions" (or any other entry type) more than once

src/cli/dev/dev-server/main.ts
```diff
@@ -174,6 +174,20 @@ export async function createDevServer(
           devLogger.log("All functions removed");
         }
       }
+      
+      if (name === "entities") {
+        const previousEntityCount = db.getCollectionNames().length;
+        db.dropAll();
```
artemdemo, line 180: Discussed this part with Dor. We want to start with clearing everything, after that we will think about introducing improvements

src/cli/dev/dev-server/database.ts
```diff
@@ -18,4 +17,11 @@ export class Database {
   getCollectionNames(): string[] {
     return Array.from(this.collections.keys());
   }
+
+  dropAll() {
+    for (const collection of this.collections.values()) {
+      collection.remove({}, { multi: true });
+    }
+    this.collections.clear();
```
kfirstri, line 25: I wonder.. do we need to clean each collection before ".clear()"ing? wont the data be removed by the clear?

src/cli/dev/dev-server/main.ts
kfirstri: I like

src/cli/dev/dev-server/database.ts
artemdemo: hm, `clear` will only clear the Map it doesn't affect map values directly. After that it's up to GC.
I just prefer to be explicit and to try to be on the safe side

PR #367, opened 2026-03-05
feat(connectors): add `connectors list-available` command

Paveltarno:
> [!NOTE]
> ## Description
>
> This PR adds a new `connectors list-available` subcommand that fetches and displays all available integration types from the Base44 API. Each integration is rendered in human-readable YAML format with a reusable `formatYaml` helper that strips empty/null values by default. The changes align with the existing connectors command pattern and include full test coverage for both the command and the YAML utility.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `connectors list-available` command that calls `GET /external-auth/available-integrations` and prints each integration as YAML
> - Added `listAvailableIntegrations()` API function in `core/resources/connector/api.ts` with Zod validation and snake_case→camelCase transform
> - Added `ListAvailableIntegrationsResponseSchema` and `ConnectionConfigFieldSchema` to `connector/schema.ts`
> - Added `formatYaml()` helper in `packages/cli/src/cli/utils/yaml.ts` with configurable empty-value stripping using `lodash.isEmpty`
> - Registered `list-available` as a subcommand on the existing `connectors` command
> - Added `yaml` package dependency and updated `bun.lock`
> - Added integration tests for all edge cases (success, empty list, API error, invalid data, missing project)
> - Added unit tests for `formatYaml` covering null/undefined/empty stripping and nested objects
> - Removed `packages/cli/plan.md` (working notes, not needed in repo)
> - Added `.claude/worktrees/` to `.gitignore`
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `AvailableIntegrationSchema` intentionally uses `z.string()` instead of `IntegrationTypeSchema` so the command works with integration types not yet defined in the CLI's enum — this ensures forward compatibility as new integrations are added on the server side.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-09 00:00 UTC</sub>


package.json
```diff
@@ -80,6 +80,7 @@
     "tmp-promise": "^3.0.3",
     "typescript": "^5.7.2",
     "vitest": "^4.0.16",
+    "yaml": "^2.8.2",
```
Paveltarno: added this yaml dep, i think it's worth it:
1. we should probably just start formatting yaml as part of our regular clack output, it integrates pretty well and we don't need to manually fiddle with formatting our domain entities
2. we might just add `--output yaml` with no extra cost

packages/cli/src/cli/commands/connectors/list-available.ts
```diff
@@ -0,0 +1,56 @@
...
+      successMessage: "Available integrations fetched successfully",
+      errorMessage: "Failed to fetch available integrations",
+    },
+  );
+
+  if (integrations.length === 0) {
+    return { outroMessage: "No available integrations found." };
+  }
+
+  for (const i of integrations) {
```
artemdemo: `i` is also usual name in `for` loops and very general, hence confusing in this case.
Maybe rename it to `integration`?

packages/cli/src/cli/commands/connectors/list-available.ts
```diff
@@ -0,0 +1,56 @@
...
+    const data: Record<string, unknown> = {
+      type: i.integrationType,
+      description: i.description,
+    };
```
artemdemo: Can't we define specific type for it and not just general one?

packages/cli/src/cli/commands/connectors/list-available.ts
```diff
@@ -0,0 +1,56 @@
...
+      data.connectionConfigFields = i.connectionConfigFields.map((f) => {
+        const field: Record<string, unknown> = {};
+        for (const [k, v] of Object.entries(f)) {
+          if (v != null) field[k] = v;
+        }
+        return field;
+      });
+    }
+    const yaml = stringify(data, { indent: 2 }).trimEnd();
+    log.info(`${i.displayName}\n  ${yaml.replace(/\n/g, "\n  ")}`);
```
artemdemo: Then maybe this approach should be wrapped in some helper function like `yamlFormatting`, so `replace` can be part of it and we wouldn't need to copy-paste it when will use it again.
so usage will be something like:
```
log.info(`${i.displayName}\n  ${yamlFormatting(data)}`);
```

packages/cli/src/cli/commands/connectors/list-available.ts
Paveltarno: @claude do your thing

kfirstri: I don't think it did anything

Paveltarno: קלוד לא חבר

Paveltarno: Fixed — renamed `i` to `integration`.

packages/cli/src/cli/commands/connectors/list-available.ts
Paveltarno: removed the type alltogether

packages/cli/src/cli/commands/connectors/list-available.ts
```diff
@@ -0,0 +1,48 @@
...
+  );
+
+  if (integrations.length === 0) {
+    return { outroMessage: "No available integrations found." };
+  }
+
+  for (const { displayName, ...rest } of integrations) {
+    const yaml = formatYaml(rest);
+    const pad = " ".repeat(YAML_INDENT);
+    log.info(`${displayName}\n${pad}${yaml.replace(/\n/g, `\n${pad}`)}`);
```
artemdemo, line 32: @Paveltarno why this part is not happening in `formatYaml`?

artemdemo: I'm referring to padding.

Paveltarno: good question, this is because the yamlformatter only formats objects into yaml
but the whole output is indented because we also have the `displayName` as a header

Screenshot 2026-03-09 at 18 49 55

this is nothing to do with the yaml formatter, more with how the CLI decides to embed that format into it's output

artemdemo: I see, thank you for clarification

PR #381, opened 2026-03-08
feat(functions): add `functions delete` command

yardend-wix:
> [!NOTE]
> ## Description
>
> Adds a `base44 functions delete <names...>` subcommand that deletes one or more deployed backend functions by name. Each function deletion runs as an individual task spinner, with a bulk summary outro for multi-name invocations. The command handles 404 responses gracefully (reports "not found" rather than erroring) and supports both space-separated variadic args and comma-separated names (e.g. `functions delete func-a,func-b`).
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `packages/cli/src/cli/commands/functions/delete.ts` — new `delete` subcommand with per-name task spinners, graceful 404 handling, comma-separated name parsing, and a bulk summary outro
> - Updated `packages/cli/src/cli/commands/functions/index.ts` — registers `getDeleteCommand` alongside `deploy` and `list` under the `functions` command group
> - Added `deleteSingleFunction(name)` to `packages/cli/src/core/resources/function/api.ts` — issues `DELETE backend-functions/:name` via the app client with `ApiError.fromHttpError()` error wrapping
> - Added `packages/cli/tests/cli/functions_delete.spec.ts` — 5 integration test cases: single delete, multi-delete summary, 404 not-found, 500 error handling, and unauthenticated/no-project failure
> - Extended `TestAPIServer` in the testkit with `mockSingleFunctionDelete()` and `mockSingleFunctionDeleteError()` mock route helpers
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `delete` command differentiates between missing functions (404) and unexpected API errors in both per-task output and the exit summary. Input validation rejects invocations with no names after parsing, and the `preAction` hook ensures the error is surfaced before the action runs.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-12 14:04 UTC</sub>

### DIFF
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -18,11 +18,6 @@ on:
         required: false
         default: false
         type: boolean
-      update_changelog:
-        description: "Update CHANGELOG.md with release notes"
-        required: false
-        default: false
-        type: boolean
       notify_skills_repo:
         description: "Notify the skills repo about the release"
         required: false
@@ -43,6 +38,7 @@ jobs:
         with:
           app-id: ${{ vars.BASE44_GITHUB_ACTIONS_APP_ID }}
           private-key: ${{ secrets.BASE44_GITHUB_ACTIONS_APP_PRIVATE_KEY }}
+          owner: base44
 
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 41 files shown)


packages/cli/src/cli/utils/parseNames.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Parse names from variadic CLI args, supporting comma-separated values.
+ * e.g. ["fn-a", "fn-b,fn-c"] → ["fn-a", "fn-b", "fn-c"]
+ */
+export function parseNames(args: string[]): string[] {
```
kfirstri: Something is no clear to me.. i see this function parses the names by comma, but in the command the description is `   .argument("<names...>", "Function names to delete")`, so do they need to pass commas or no?
Also i don't tihnk this should be in the cli/utils but inside the function or something

packages/cli/src/cli/commands/functions/delete.ts
```diff
@@ -0,0 +1,73 @@
...
+    .argument("<names...>", "Function names to delete")
+    .action(async (rawNames: string[]) => {
+      await runCommand(
+        () => {
+          const names = parseNames(rawNames);
+          if (names.length === 0) {
+            throw new InvalidInputError(
+              "At least one function name is required",
+            );
+          }
```
kfirstri: Can we use the .preAction hook and move this to a validation function?

packages/cli/src/cli/commands/functions/delete.ts
```diff
@@ -0,0 +1,73 @@
...
+        log.warn(`${name.padEnd(25)} not found`);
+        notFound++;
+      } else {
+        log.error(
+          `${name.padEnd(25)} error: ${error instanceof Error ? error.message : String(error)}`,
+        );
+        errors++;
+      }
+    }
+    completed++;
```
kfirstri: How about we simplify this function to just get 1 function name and delete it? or is it weird?

anyway for this - we need to use `runTask` so we will have a spinner, then for each we can show a better message

packages/cli/src/cli/utils/parseNames.ts
yardend-wix: Done — moved `parseNames` inline into `delete.ts` and removed the separate utility file. The comma support is just a convenience on top of Commander variadic args (e.g. `delete func-a,func-b` works the same as `delete func-a func-b`).

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

packages/cli/src/cli/commands/functions/delete.ts
yardend-wix: Done — added a `validateNames` function with `.hook("preAction", validateNames)`, following the same pattern used in `link.ts` and `create.ts`.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

packages/cli/src/cli/commands/functions/delete.ts
yardend-wix: Kept it supporting multiple names since Commander variadic `<names...>` naturally handles both cases, and the single-name path already has a clean output. Wrapped each deletion in `runTask` so there's a spinner per function with success/error messages.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

kfirstri (approved):

PR #395, opened 2026-03-09
Schemas validation while running `base44 dev`

artemdemo:
> [!NOTE]
> ## Description
>
> This PR adds entity record validation to the dev server, introducing a `Validator` class that enforces entity schemas on create and update operations. When a record fails validation (wrong types, missing required fields), the API now returns a `422 Unprocessable Entity` response with a structured error object — matching the shape of production API validation errors. The `Database` class is also reorganized into a `db/` subdirectory to better group schema-aware logic.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `packages/cli/src/cli/dev/dev-server/db/validator.ts` — new `Validator` class with `filterFields()`, `applyDefaults()`, and `validate()` methods supporting all Base44 field types (`string`, `integer`, `number`, `boolean`, `array`, `object`), required-field checks, partial validation, and recursive nested type validation
> - Added `packages/cli/src/cli/dev/dev-server/db/database.ts` — refactored `Database` class (moved from `database.ts`) that now stores entity schemas and exposes `validate()` and `prepareRecord()` methods
> - Deleted `packages/cli/src/cli/dev/dev-server/database.ts` — replaced by `db/database.ts`
> - Updated `routes/entities.ts` — POST (single and bulk) and PATCH handlers now call `db.prepareRecord()` + `db.validate()` before inserting/updating; invalid records return `422` with a `ValidationError` body
> - Updated `core/resources/entity/schema.ts` — exported `PropertyDefinition` type, made `type` field required (was optional), and cleaned up recursive Zod type annotations
> - Added `packages/cli/tests/scripts/validator.spec.ts` — comprehensive unit test suite for the `Validator` class covering all field types, required fields, partial validation, nested arrays/objects, defaults, and error shapes
> - Minor: fixed `vitest.config.ts` to use `node:path` import; added `biome-ignore` comment in connector test
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The validator intentionally silently ignores unknown fields (relying on `filterFields()` to strip them beforehand) and skips `items`/`properties` sub-validation when those schema keys are absent, with inline test comments noting this matches current production behavior.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-11 00:00 UTC</sub>

### DIFF
--- a/packages/cli/src/cli/dev/dev-server/db/validator.ts
+++ b/packages/cli/src/cli/dev/dev-server/db/validator.ts
@@ -129,7 +129,7 @@ export class Validator {
       return {
         hasError: true,
         error: this.createValidationError(
-          `Error in field ${fieldPath}: Input should be a valid ${propertyType}`,
+          `Error in field ${fieldPath}: Unsupported field type ${propertyType}`,
         ),
       };
     }
--- a/packages/cli/tests/scripts/validator.spec.ts
+++ b/packages/cli/tests/scripts/validator.spec.ts
@@ -351,7 +351,7 @@ describe("Validator", () => {
 
         expect(result.hasError).toBe(true);
         if (result.hasError) {
-          expect(result.error.message).toContain("date");
+          expect(result.error.message).toContain("Unsupported field type date");
         }
       });
     });


packages/cli/vitest.config.ts
```diff
@@ -1,5 +1,5 @@
+import { resolve } from "node:path";
 import { defineConfig } from "vitest/config";
-import { resolve } from "path";
```
artemdemo, line 2: not related to PR, just cleaning up some linting

packages/cli/src/cli/dev/dev-server/database.ts
artemdemo, line 1: Just moved to a new location, inside `db` directory

packages/cli/src/core/resources/entity/schema.ts
```diff
@@ -106,8 +106,8 @@ const FieldRLSSchema = z.looseObject({
   delete: RLSRuleSchema.optional(),
 });
 
-const PropertyDefinitionSchema: z.ZodType<unknown> = z.looseObject({
```
artemdemo, line 109: I don't see reason for it to be `unknown` when types are casted.
(same below)
Am i I missing something?

kfirstri: I think you're right

packages/cli/src/cli/dev/dev-server/db/validator.ts
```diff
@@ -0,0 +1,230 @@
...
+const fieldTypes = [
+  "string",
+  "integer",
+  "number",
+  "boolean",
+  "array",
+  "object",
+];
+
+export class Validator {
```
kfirstri, line 34: This is a high level idea, haven't dived into it - 
But as far as i understand the entities schema are basically json-schema standard, and i saw this package @cfworker/json-schema (https://github.com/cfworker/cfworker) which we can feed the schema to and then just call validate?

artemdemo: The thing is that leaders don't really want to validate - this [schema is just documentation of intent](https://base44workspace.slack.com/archives/C0A6NUPQXKJ/p1773238212713629?thread_ts=1773227693.561529&cid=C0A6NUPQXKJ)
And looks we're [going with this approach forward](https://base44workspace.slack.com/archives/C0A6NUPQXKJ/p1773253737657309?thread_ts=1773227693.561529&cid=C0A6NUPQXKJ).
`json-schema` supports far more features than what we need to handle.
Therefore not sure if need to introduce anything that will actually use this schema to validate any further than just types.

artemdemo: Looked into it once more.
It's possible to use `json-schema`, not sure if makes code simpler though.
I will still need to do recursions in order to make `json-schema` work with limited set of validations.
And also I will need to filter out errors that are not relevant to our case.

If you want to take a look, here is a prompt that I used to test it out:
```
Take a look at how I validate using a schema. Basically, I am only interested in the property type and whether it's required or not. I don't validate other constraints like maxLength, pattern, minLength, etc. I do validate recursively, though. If a property is an object or an array, I will go and check its properties/items, but will only validate the same two things: type and whether it's required. So, I'm thinking whether I can use `@cfworker/json-schema` instead for the same thing. This package does much more, but maybe I can silence other errors or something.
```

kfirstri: I think we should proceed with the current implentation considering the links you added, I just don't love maintaining this validation logic (and issues that may raise) on the CLI side.. but it looks kinda simple for now 👍

artemdemo: we will need anyway maintain validation logic in some capacity 😞

PR #406, opened 2026-03-11
User Entity Schema

artemdemo:
> [!NOTE]
> ## Description
>
> This PR adds local User entity support to the `base44 dev` server, replacing the previous behavior of proxying all `/User` requests to production. The current authenticated user is automatically seeded into an in-memory User collection on startup, and a new JWT-aware router handles CRUD operations for the User entity locally. Entity names are now normalized to lowercase for consistent lookup.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added a dedicated `entities-user-router.ts` that handles `GET /User/:id` (including `me`), `POST /User/`, `PUT /User/:id` with JWT-based auth via a `withAuth` wrapper
> - `Database.load()` is now async and automatically seeds a `User` collection with the current authenticated user's info (from `readAuth()`)
> - Added `buildUserSchema()` to merge built-in User fields (`full_name`, `email`) with any custom User entity schema, and validates that built-in fields are not overridden
> - Refactored `entities.ts` → `routes/entities/entities-router.ts` and removed the `remoteProxy` parameter (no longer needed for User routes)
> - Moved `stripInternalFields` utility to `utils.ts` and entity name normalization (`.toLowerCase()`) added to `Database`
> - Changed `Validator` to throw `EntityValidationError` instead of returning a validation response object; callers updated to catch and return HTTP 422
> - Added `jsonwebtoken` dependency for decoding Bearer tokens in the local dev server
> - Updated validator tests to match the new throw-based validation API
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The User collection is pre-populated with the logged-in developer's info (email, name, role=admin) so that local apps behave similarly to production when reading the current user. Restricted fields (`full_name`, `email`, `role`) cannot be updated via the entities API, mirroring production backend constraints. The `POST /User/bulk` endpoint is a no-op, consistent with production behavior.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-22 00:00 UTC</sub>

### DIFF
--- a/packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
+++ b/packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
@@ -1,8 +1,9 @@
+import type { Document } from "@seald-io/nedb";
 import type { Request, Response, Router } from "express";
 import { Router as createRouter, json } from "express";
+import jwt from "jsonwebtoken";
 import { nanoid } from "nanoid";
 import type { Logger } from "@/cli/dev/createDevLogger.js";
-import { readAuth } from "@/core/index.js";
 import { type Database, USER_COLLECTION } from "../../db/database.js";
 import {
   type EntityRecord,
@@ -10,55 +11,78 @@ import {
 } from "../../db/validator.js";
 import { getNowISOTimestamp, stripInternalFields } from "../../utils.js";
 
+type UserDocument = Document<{
+  email: string;
+  id: string;
+}>;
+
 export function createUserRouter(db: Database, logger: Logger): Router {
   const router = createRouter({ mergeParams: true });
   const parseBody = json();
 
-  router.use((req, res, next) => {
-    const auth = req.headers.authorization;
-    if (!auth || !auth.startsWith("Bearer ")) {
-      res.status(401).json({ error: "Unauthorized" });
-      return;
-    }
-    next();
-  });
+  function withAuth<R>(
+    handler: (
+      req: Request<R>,
+      res: Response,
+      currentUser: UserDocument,
+    ) => Promise<void> | void,
+  ): (req: Request<R>, res: Response) => Promise<void> {
+    return async (req, res) => {
+      const auth = req.headers.authorization;
+      if (!auth || !auth.startsWith("Bearer ")) {
+        res.status(401).json({ error: "Unauthorized" });
+        return;
+      }
+      try {
+        const { payload } =
+          jwt.decode(auth.replace("Bearer ", ""), { complete: true }) ?? {};
 
-  router.get("/:id", async (req: Request<{ id: string }>, res: Response) => {
-    let result: Record<string, unknown> | undefined;
+        const result = await db
+          .getCollection(USER_COLLECTION)
+          ?.findOneAsync({ email: payload?.sub });
 
-    if (req.params.id === "me") {
-      const userInfo = await readAuth();
-      result = await db
+        if (!result) {
+          res
+            .status(404)
+            .json({ error: "Unable to read data for the current user" });
+          return;
+        }
+
+        await handler(req, res, result as UserDocument);
+      } catch {
+        res.status(401).json({ error: "Unauthorized" });
+      }
+    };
+  }
+
+  router.get(
+    "/:id",
+    withAuth<{ id: string }>(async (req, res, currentUser) => {
+      const id = req.params.id === "me" ? currentUser.id : req.params.id;
+      const result = await db
         .getCollection(USER_COLLECTION)
-        ?.findOneAsync({ email: userInfo.email });
-    } else {
-      result = await db
-        .getCollection(USER_COLLECTION)
-        ?.findOneAsync({ id: req.params.id });
-    }
+        ?.findOneAsync({ id });
 
-    if (!result) {
-      res
-        .status(404)
-        .json({ error: `User with id "${req.params.id}" not found` });
-      return;
-    }
-    res.json(stripInternalFields(result));
-  });
+      if (!result) {
+        res
+          .status(404)
+          .json({ error: `User with id "${req.params.id}" not found` });
+        return;
+      }
+      res.json(stripInternalFields(result));
+    }),
+  );
 
-  router.post("/", parseBody, async (req, res) => {
-    const userInfo = await readAuth();
-    const currentUser = await db
-      .getCollection(USER_COLLECTION)
-      ?.findOneAsync({ email: userInfo.email });
-
-    if (currentUser) {
+  router.post(
+    "/",
+    parseBody,
+    withAuth(async (req, res, currentUser) => {
       const now = getNowISOTimestamp();
 
       // Production is not allowing to add user entity directly.
       // In case developer tries to do it - backend silently fails.
       res.json({
-        created_by: userInfo.email,
+        created_by: currentUser.email,
         created_by_id: currentUser.id,
         id: nanoid(),
         created_date: now,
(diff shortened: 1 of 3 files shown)


packages/cli/src/cli/utils/command/Base44Command.ts
```diff
@@ -108,6 +108,7 @@ export class Base44Command extends Command {
   /** @public - called by Commander internally via command dispatch */
   override action(
     // biome-ignore lint/suspicious/noExplicitAny: must match Commander.js action() signature
+    // biome-ignore lint/suspicious/noConfusingVoidType: must match Commander.js action() signature
```
artemdemo, line 111: Not related to my PR, but started to fail for me.
Ignoring

packages/cli/tests/scripts/validator.spec.ts
```diff
@@ -89,126 +90,138 @@ describe("Validator", () => {
     it("should pass when all required fields are present", () => {
       const schema = makeEntity({ name: { type: "string" } }, ["name"]);
 
-      const result = validator.validate({ name: "Alice" }, schema);
-
-      expect(result.hasError).toBe(false);
+      expect(() => validator.validate({ name: "Alice" }, schema)).not.toThrow();
```
artemdemo, line 93: Validator is now throwing an error, instead of returning object that describes the error.
Updating tests to handle it.
Here and below.

packages/cli/src/cli/dev/dev-server/db/validator.ts
```diff
@@ -74,19 +80,16 @@ export class Validator {
         entitySchema,
       );
       if (requiredFieldsResponse.hasError) {
-        return requiredFieldsResponse;
+        throw new EntityValidationError(requiredFieldsResponse.error);
```
artemdemo, line 83: I found that it's more convenient to throw the error, rather than return an object that describes the error.
So I decided to refactor it a bit.

packages/cli/src/cli/dev/dev-server/db/database.ts
```diff
@@ -55,4 +121,8 @@ export class Database {
     }
     return this.validator.applyDefaults(filteredRecord, schema);
   }
+
+  private normalizeName(entityName: string): string {
+    return entityName.toLowerCase();
+  }
```
artemdemo, line 127: Behind the scenes production is treating entity name as case-insensitive: "User" and "user" are the same.

packages/cli/src/cli/dev/dev-server/main.ts
```diff
@@ -76,6 +76,15 @@ export async function createDevServer(
...
   });
 
+  app.use((req, res, next) => {
+    const auth = req.headers.authorization;
+    if (!auth || !auth.startsWith("Bearer ")) {
+      res.status(401).json({ error: "Unauthorized" });
+      return;
+    }
+    next();
+  });
```
kfirstri: So every request to the dev-server must have an authorization header. Is this also how the apper/ backend works? It won't allow anonymous calls at all?

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export function createUserRouter(db: Database, logger: Logger): Router {
+  const router = createRouter({ mergeParams: true });
+  const parseBody = json();
+
+  router.get("/:id", async (req: Request<{ id: string }>, res: Response) => {
+    let result: Record<string, unknown> | undefined;
+
+    if (req.params.id === "me") {
+      const userInfo = await readAuth();
```
kfirstri: Wont the request at this point will have a authorization token that we can read some user data from? then we don't need the readAuth call

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+    const userInfo = await readAuth();
+    const currentUser = await db
+      .getCollection(USER_COLLECTION)
+      ?.findOneAsync({ email: userInfo.email });
+
+    if (currentUser) {
+      const now = getNowISOTimestamp();
+
+      // Production is not allowing to add user entity directly.
+      // In case developer tries to do it - backend silently fails.
```
kfirstri, line 83: I don't understand this whole POST / endpoint .. we're validating the developing (since we're using readAuth), and if we found the developer account we just return an object, but not really save it? so basically it's just a mock.. but production also does not allow it?

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+      res
+        .status(404)
+        .json({ error: "Unable to read data for the current user" });
+    }
+  });
+
+  router.post("/bulk", async (_req, res) => {
+    // not supported in direct call: NO-OP
+    res.json({});
+  });
```
kfirstri, line 99: Is this a noop endpoint because we just don't want to support it? can't we just delete it and it will return 404?

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  router.put("/:id", parseBody, async (req: Request<{ id: string }>, res) => {
+    // User is only allowed to update himself, not other users
+    if (req.params.id === "me") {
+      const userInfo = await readAuth();
+      const collection = db.getCollection(USER_COLLECTION);
+      const userRecord = await collection?.findOneAsync({
+        email: userInfo.email,
+      });
```
kfirstri: If this is really how we will get the developer account (raising it again - do we have it as an auth header maybe?), then this is the 3rd time we're doing this in this file, so maybe refactor to another function

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+          }
+
+          res.json(stripInternalFields(updResult.affectedDocuments));
+        } catch (error) {
+          if (error instanceof EntityValidationError) {
+            res.status(422).json(error.context);
+            return;
+          }
+          logger.error(
+            `Error in PUT /${USER_COLLECTION}/${req.params.id}:`,
```
kfirstri: This error looks weird, no need to use USER_COLLECTION inside the url that we're printing

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+            error,
+          );
+          res.status(500).json({ error: "Internal server error" });
+        }
+      } else {
+        res.status(404).json({ error: `User record not found` });
+      }
+    } else {
+      res.json({});
```
kfirstri: IF the user can only update himself and not others, shouldn't we return some errors? this is how apper behaves? :S

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+  });
+
+  router.delete("/:id", async (_req, res) => {
+    // not supported in direct call: NO-OP
+    res.json({});
+  });
```
kfirstri, line 160: same question like in /bulk route, maybe we can just remove it? what does "direct call" mean?

packages/cli/src/cli/dev/dev-server/db/database.ts
```diff
@@ -1,21 +1,87 @@
...
+    const builtInFields = {
+      full_name: { type: "string" as const },
+      email: { type: "string" as const },
+    };
+
+    if (!customUserEntity) {
+      return {
+        name: "User",
+        type: "object",
+        properties: { ...builtInFields, role: { type: "string" } },
```
kfirstri, line 65: "role" is also a built-in field https://docs.base44.com/developers/backend/resources/entities/user-schema#built-in-fields

artemdemo: Yes, I know, but this field is not validated like `full_name` and `email` and error is not thrown when user tries to override it

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: This is exactly how productions behaves.
It returns updated object, while not updating Entries list.

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: This is not how production behaves - in production it's no-op

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: Why? Can you elaborate, what is weird about it?
Can you give example of an error that you have in mind?

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: https://github.com/base44/cli/pull/406/changes#r2971255264

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: Right. Also `admin` is able to update any user. So I will remove this part of the statement.

packages/cli/src/cli/dev/dev-server/main.ts
artemdemo: Right, this is overkill here. Currently it's needed only for `/User`

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: updated

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
kfirstri: It's probably fine for now, but i mean that it's just a coincidence that the endpoint route /users/:id has the same name as the collection, the collection name can be "users_collection", but the route will remain "users/"

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -0,0 +1,163 @@
...
+type UserDocument = Document<{
+  email: string;
+  id: string;
+}>;
+
+export function createUserRouter(db: Database, logger: Logger): Router {
+  const router = createRouter({ mergeParams: true });
+  const parseBody = json();
+
+  function withAuth<R>(
```
kfirstri, line 23: love this

kfirstri (approved):
noice

PR #422, opened 2026-03-17
feat: add auth password-login, pull, and push commands

moranshe-max:
> [!NOTE]
> ## Description
>
> Adds a new `base44 auth` command group for managing app authentication configuration. The feature introduces three subcommands (`password-login`, `pull`, `push`) backed by a new `auth-config` core resource module that handles reading/writing local auth config files and syncing them with the Base44 API. Auth config is also integrated into the existing `base44 deploy` pipeline so it is pushed automatically alongside entities, functions, agents, and connectors.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **New `auth` command group** (`packages/cli/src/cli/commands/auth/`):
>   - `auth password-login <enable|disable>` — toggle username/password authentication in the local auth config file; warns if disabling would leave no login methods enabled
>   - `auth pull` — fetch auth config from the Base44 API and write it to the local `auth/` directory
>   - `auth push` — push the local auth config file to the Base44 API, with an interactive confirmation prompt (skippable via `--yes`)
>   - Command group registered as hidden in `program.ts`
> - **New `auth-config` core resource** (`packages/cli/src/core/resources/auth-config/`):
>   - Zod schemas for API response (snake_case → camelCase transform) and local file format
>   - `hasAnyLoginMethod()` utility and `toAuthConfigPayload()` serializer
>   - `Resource<T>`-compatible module for unified read/push lifecycle
>   - Local config read/write with JSON5 support
> - **Deploy integration** — `authDir` added to project config schema (default `"auth"`), `authConfig` included in `ProjectData`, read in parallel during `readProjectConfig()`, pushed during `deployAll()`, and shown in the deploy summary
> - **Tests** — CLI integration tests for `auth password-login` and core unit tests covering API pull/push, local file I/O, and schema validation
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `auth` command group is currently hidden from top-level help (registered with `{ hidden: true }`) while the feature is in early rollout. The `authDir` project config field defaults to `"auth"` so existing projects don't require any config changes to pick it up.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-25 10:48 UTC</sub>

### DIFF
--- a/packages/cli/src/cli/program.ts
+++ b/packages/cli/src/cli/program.ts
@@ -71,7 +71,7 @@ export function createProgram(context: CLIContext): Command {
   program.addCommand(getSecretsCommand());
 
   // Register auth config commands
-  program.addCommand(getAuthCommand());
+  program.addCommand(getAuthCommand(), { hidden: true });
 
   // Register site commands
   program.addCommand(getSiteCommand());


packages/cli/src/core/resources/auth-config/api.ts
```diff
@@ -0,0 +1,74 @@
...
+    );
+  }
+
+  return result.data.authConfig;
+}
+
+/**
+ * Returns true if at least one login method is enabled in the given config.
+ */
+export function hasAnyLoginMethod(config: AuthConfig): boolean {
```
kfirstri: I think this should be in config / schema or somewhere else because it's not really API.ts file

packages/cli/src/core/resources/auth-config/api.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Fetches the current auth config for the app.
+ */
+export async function getAuthConfig(): Promise<AuthConfig> {
+  const { id } = getAppConfig();
+
+  let response: KyResponse;
+  try {
+    response = await base44Client.get(`api/apps/${id}`);
```
kfirstri, line 16: I wonder if this should really big in project/api.ts file since it's the entire project data.. but maybe it's a future refactor

packages/cli/src/core/resources/auth-config/config.ts
```diff
@@ -0,0 +1,91 @@
...
+
+function getAuthConfigPath(authDir: string): string {
+  return join(authDir, AUTH_CONFIG_FILENAME);
+}
+
+/**
+ * Reads the auth config file from the given directory.
+ * Returns [config] if the file exists, or [] if the directory or file doesn't exist.
+ */
+export async function readAuthConfig(authDir: string): Promise<AuthConfig[]> {
```
kfirstri: why are we returning an array of AuthConfig instead just the AuthConfig? is it possible to have multiple AuthConfigs? 🤔

packages/cli/src/core/resources/auth-config/config.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Updates the auth config file with a partial update.
+ * If the file doesn't exist, creates it with default values and applies the updates.
+ */
+export async function updateAuthConfigFile(
+  authDir: string,
+  updates: Partial<AuthConfig>,
+): Promise<AuthConfig> {
```
kfirstri: I wonder if we need a read / write / update functions, maybe just read and write? write will also do updates or updates will happen before calling the config.ts?

packages/cli/src/core/resources/auth-config/push.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Pushes the auth config to the remote API.
+ * If the array is empty, does nothing.
+ */
+export async function pushAuthConfig(configs: AuthConfig[]): Promise<void> {
+  if (configs.length === 0) {
+    return;
+  }
+
+  await pushAuthConfigToApi(configs[0]);
```
kfirstri: I commented it in the config.ts file, can AuthConfig be just singular and not an array? then we don't need to keep putting into an array and out of an array 💪

packages/cli/src/cli/commands/auth/password-login.ts
```diff
@@ -0,0 +1,87 @@
...
+  return {
+    outroMessage: `Username & password authentication ${newStatus} in local config. Run \`base44 auth push\` or \`base44 deploy\` to apply.`,
+  };
+}
+
+export function getPasswordLoginCommand(): Command {
+  return new Base44Command("password-login")
+    .description("Enable or disable username & password authentication")
+    .option("--enable", "Enable password authentication")
+    .option("--disable", "Disable password authentication")
```
kfirstri: what do you think about using commander's .choices instead of 2 different flags? and then the command will look like `base44 auth password-login enable`

```
// something like
.argument('<state>').choices(['enable', 'disable'])
```

packages/cli/src/cli/commands/auth/password-login.ts
```diff
@@ -0,0 +1,87 @@
...
+  validateOptions(options);
+
+  const shouldEnable = !!options.enable;
+  const { project } = await readProjectConfig();
+
+  const configDir = dirname(project.configPath);
+  const authDir = join(configDir, project.authDir);
+
+  const updated = await runTask(
+    `${shouldEnable ? "Enabling" : "Disabling"} username & password authentication`,
```
kfirstri: maybe the message should be something like "Updating authenticaiton configuration" or something? so make sure it's only locally

kfirstri: i just saw the outro message, so maybe it's fine

packages/cli/src/cli/commands/auth/password-login.ts
```diff
@@ -0,0 +1,87 @@
...
+    async () => {
+      return await updateAuthConfigFile(authDir, {
+        enableUsernamePassword: shouldEnable,
+      });
```
kfirstri: I also commented baout this inside the updateAuthConfigFile function, i just feel writeAuthConfigFile can do everything?

packages/cli/src/cli/commands/auth/pull.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  return {
+    outroMessage: `Pulled auth config to ${authDir}`,
+  };
+}
+
+export function getAuthPullCommand(): Command {
+  return new Base44Command("pull")
+    .description("Pull auth config from Base44 to local file")
```
kfirstri, line 52: Might worth adding some text about overwriting existing file

packages/cli/src/cli/commands/auth/push.ts
```diff
@@ -0,0 +1,39 @@
...
+  );
+
+  return {
+    outroMessage: "Auth config pushed to Base44",
+  };
+}
+
+export function getAuthPushCommand(): Command {
+  return new Base44Command("push")
+    .description("Push local auth config to Base44")
```
kfirstri, line 68: product question - 
1. I saw you added a warning in "password-login" command about that users will be locked out.. maybe we should do it here too?
2. I know it's currently confusing in the CLI, but do we want to confirm this with the user first? (and add a --yes flag or --force or something?)

packages/cli/src/cli/commands/auth/password-login.ts
moranshe-max: Great tip! Claude claims it's easier for it to consume the cli that way.

packages/cli/src/cli/commands/auth/push.ts
moranshe-max: Went with --yes to keep it consistent with base44 deploy

packages/cli/src/core/project/deploy.ts
```diff
@@ -59,14 +69,18 @@ export async function deployAll(
...
+  const { project, entities, functions, agents, connectors, authConfig } =
+    projectData;
 
   await entityResource.push(entities);
   await deployFunctionsSequentially(functions, {
     onStart: options?.onFunctionStart,
     onResult: options?.onFunctionResult,
   });
   await agentResource.push(agents);
+  if (authConfig) {
```
kfirstri: 1. You should have a resource.ts file inside resources/auth-config/resource.ts with the "readAll" and "push" command configured, this is basically our interface for resourced, i don't want deploy.ts to be too familiar with resources implementation
2. move the if(authConfig) check to be inside the resource push function, like all other resourced do

PR #425, opened 2026-03-17
refactor logger to simple logger

kfirstri:
> [!NOTE]
> ## Description
>
> This PR extracts a `@base44-cli/logger` package that provides a `Logger` interface with two implementations: `ClackLogger` for interactive TTY sessions (delegates to `@clack/prompts`) and `SimpleLogger` for non-interactive/CI environments (writes plain text to stderr). The logger is injected into `CLIContext` and passed as the first argument to all action functions via `Base44Command`, replacing direct `@clack/prompts` `log` imports throughout commands.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Created `packages/logger` workspace package (`@base44-cli/logger`) with a `Logger` interface, `ClackLogger` (interactive), and `SimpleLogger` (CI/non-interactive)
> - Added `log: Logger` to `CLIContext` — instantiated as `ClackLogger` or `SimpleLogger` based on TTY detection at startup in `runCLI()`
> - Updated `Base44Command.action()` to inject `CLIContext` as the **first argument** to every action function, removing the need for commands to access it via the Commander command instance
> - Removed the `isNonInteractive` getter from `Base44Command` (now accessed via destructured context)
> - Updated `ensureAuth` and `ensureAppConfig` middleware to accept `CLIContext` instead of just `ErrorReporter`
> - Migrated all command files to destructure `{ log }` / `{ log, isNonInteractive }` from the injected context instead of importing `log` from `@clack/prompts` directly
> - Updated `docs/commands.md` to reflect the new context-injection pattern and `Logger` usage
> - Re-exported `Logger`, `ClackLogger`, and `SimpleLogger` from `@/cli/utils/index.js`
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `Logger` interface maps 1:1 to `@clack/prompts` log methods (`info`, `success`, `warn`, `error`, `step`, `message`), so migrating commands is mechanical. The `SimpleLogger` writes all output to stderr to keep stdout clean for machine-readable output in CI pipelines.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-23 00:00 UTC</sub>

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -5,7 +5,7 @@
     "packages/cli": {
       "entry": ["src/cli/index.ts", "bin/binary-entry.ts", "tests/**/testkit/index.ts"],
       "project": ["src/**/*.ts", "tests/**/*.ts"],
-      "ignore": ["dist/**", "tests/fixtures/**"],
+      "ignore": ["tests/fixtures/**"],
       "ignoreDependencies": ["@types/deno"]
     },
     "packages/logger": {
--- a/packages/logger/package.json
+++ b/packages/logger/package.json
@@ -3,7 +3,6 @@
   "version": "0.0.1",
   "description": "Logger interface and implementations for Base44 CLI",
   "type": "module",
-  "main": "./src/index.ts",
   "exports": {
     ".": "./src/index.ts"
   },
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -3,7 +3,7 @@
     "target": "ES2022",
     "module": "ES2022",
     "lib": ["ES2022"],
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,


packages/logger/src/ClackLogger.ts
artemdemo, line 1: Maybe `logger` should be a separate package?
wdyt?

kfirstri: Love this idea, what do you think about the name @base44-cli/logger? maybe just @base44/logger?

packages/logger/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@base44-cli/logger",
+  "version": "0.0.1",
+  "description": "Logger interface and implementations for Base44 CLI",
+  "type": "module",
+  "main": "./src/index.ts",
```
artemdemo: Why do you need `main`?

knip.json
```diff
@@ -7,6 +7,9 @@
       "project": ["src/**/*.ts", "tests/**/*.ts"],
       "ignore": ["dist/**", "tests/fixtures/**"],
       "ignoreDependencies": ["@types/deno"]
+    },
+    "packages/logger": {
+      "project": ["src/**/*.ts"]
```
artemdemo, line 12: Here you don't need `entry`?
Why it's needed in `packages/cli`?

knip.json
```diff
@@ -7,6 +7,9 @@
       "project": ["src/**/*.ts", "tests/**/*.ts"],
       "ignore": ["dist/**", "tests/fixtures/**"],
```
artemdemo: Do we really need `"dist/**"` in `ignore`?

kfirstri: good catch, removed

knip.json
kfirstri: looked it up, knip recognize the entry by the exports, and since the cli package uses the bin/ folder we need to specify it.. but i removed it for logger package.

packages/logger/package.json
kfirstri: so our tsconfig was set to "moduleResolution": "node" which didn't support "exports", only "main".. working on this

tsconfig.base.json
```diff
@@ -3,7 +3,7 @@
     "target": "ES2022",
     "module": "ES2022",
     "lib": ["ES2022"],
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
kfirstri, line 6: @artemdemo what do you think about this?
this makes it so packag.json will support "exports" and not just "main"
in wix-cli we use "node16", but it seems it raises other issues.. maybe i can first bump to node16 in another branch and then fix this

artemdemo: yes, I think we can do it

PR #441, opened 2026-03-24
non interactive runTask function

kfirstri:
> [!NOTE]
> ## Description
>
> This PR refactors `runTask` from a standalone exported function into a dependency-injected factory pattern. Instead of importing `runTask` directly, commands now receive it via `CLIContext`, with two implementations: `createInteractiveRunTask` (uses clack spinner for TTY) and `createSimpleRunTask` (uses plain log output for non-interactive/CI environments). This ensures that spinner-based UI is never used in non-TTY contexts.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Replaced the exported `runTask` function with `createInteractiveRunTask()` and `createSimpleRunTask(log)` factory functions in `runTask.ts`
> - Added `RunTaskFn` type alias exported from `runTask.ts`
> - Added `runTask: RunTaskFn` field to `CLIContext` interface in `types.ts`
> - Instantiated the correct `runTask` implementation in `index.ts` based on `isNonInteractive` flag, and passed it into `CLIContext`
> - Removed `runTask` from the `cli/utils/index.ts` barrel export (no longer a free-standing import)
> - Updated all command files (~20 files) to destructure `runTask` from `CLIContext` instead of importing it directly
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `createSimpleRunTask` implementation logs the start message via `log.info`, update messages via `log.info`, success via `log.success`, and errors via `log.error` — making non-interactive output readable in CI/piped environments without spinner control codes.
>
> ---
> <sub>🤖 Generated by Claude | 2026-03-29 00:00 UTC</sub>

### DIFF
--- a/packages/cli/src/cli/utils/runTask.ts
+++ b/packages/cli/src/cli/utils/runTask.ts
@@ -31,9 +31,6 @@ export function createInteractiveRunTask(): RunTaskFn {
   };
 }
 
-/**
- * Creates a RunTaskFn that delegates to the provided logger for non-interactive environments.
- */
 export function createSimpleRunTask(log: Logger): RunTaskFn {
   return async (startMessage, operation, options) => {
     log.info(startMessage);
--- a/packages/cli/src/cli/commands/auth/password-login.ts
+++ b/packages/cli/src/cli/commands/auth/password-login.ts
@@ -1,7 +1,7 @@
 import { dirname, join } from "node:path";
 import type { Command } from "commander";
 import type { CLIContext, RunCommandResult } from "@/cli/types.js";
-import { Base44Command, runTask } from "@/cli/utils/index.js";
+import { Base44Command } from "@/cli/utils/index.js";
 import { InvalidInputError } from "@/core/errors.js";
 import { readProjectConfig } from "@/core/project/index.js";
 import {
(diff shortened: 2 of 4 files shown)


packages/cli/src/cli/utils/runTask.ts
```diff
@@ -1,62 +1,52 @@
...
+      return result;
+    } catch (error) {
+      s.error(options?.errorMessage || "Failed");
+      throw error;
+    }
+  };
+}
+
+/**
+ * Creates a RunTaskFn that delegates to the provided logger for non-interactive environments.
```
artemdemo: Delegates what? :)
Description is a bit unclear

packages/cli/src/cli/utils/runTask.ts
artemdemo, line 1: > @artemdemo I wonder what do you think about this. moving the runTask to be injected from the CLIContext.
> It's nice to have it support non interactive and such
> I feel that maybe it's not the right place for runTask 🤔

non interractive `runTask` is a good idea.

but what do you mean about the place? Like you don't want to place it in the context?
You're actually don't have to.
You can just check interactivity in `runTask` itself and then use spinner or logger

PR #445, opened 2026-03-26
feat: add social login and custom Google OAuth commands

moranshe-max:
> [!NOTE]
> ## Description
>
> Adds a new `base44 auth social-login <provider> <enable|disable>` CLI command for managing social login providers (Google, Microsoft, Facebook, Apple). Google additionally supports custom OAuth via `--client-id`, `--client-secret`, `--client-secret-stdin`, and `--env-file` options to push credentials directly to the backend secrets API. This PR also introduces two reusable utilities — `resolveSecret` for multi-source secret collection (flag → stdin → env var → interactive prompt) and `readStdin` for safely reading piped input.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `packages/cli/src/cli/commands/auth/social-login.ts`: new `social-login` CLI command with provider/action arguments and Google-specific custom OAuth flags
> - Added `packages/cli/src/core/resources/auth-config/social-login.ts`: core functions `updateSocialLoginConfig` and `pushCustomOAuthSecret` for reading/writing local auth config and pushing secrets to the API
> - Added `packages/cli/src/core/resources/auth-config/schema.ts`: `SOCIAL_PROVIDERS` registry and `ProviderName` type with per-provider schema definitions (toggle field, optional custom OAuth fields)
> - Added `packages/cli/src/cli/utils/secret-input.ts`: reusable `resolveSecret` utility with precedence chain (stdin > flag > env var > interactive prompt)
> - Added `packages/cli/src/cli/utils/stdin.ts`: `readStdin` utility for safely consuming piped stdin input
> - Wired `social-login` into the `auth` command group (`auth/index.ts`) and exported new utilities from `cli/utils/index.ts`
> - Refactored `password-login.ts` to use Commander's `Argument.choices()` instead of a manual runtime validation function
> - Added comprehensive integration tests (`tests/cli/auth_social_login.spec.ts`, `tests/cli/secret_input.spec.ts`) covering argument validation, provider restrictions, happy paths, and custom OAuth flows
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `resolveSecret` and `readStdin` utilities are designed for reuse by any future command that needs to accept a secret from multiple input sources (flag, stdin pipe, env var, or interactive prompt). The `SOCIAL_PROVIDERS` registry in `schema.ts` makes it straightforward to add new providers or extend Google-style custom OAuth support to other providers in the future.
>
> ---
> <sub>🤖 Generated by Claude | 2026-04-15 00:00 UTC | 7d24dfc</sub>


packages/cli/src/cli/commands/auth/password-login.ts
```diff
@@ -68,6 +42,11 @@ async function passwordLoginAction(
 export function getPasswordLoginCommand(): Command {
   return new Base44Command("password-login")
     .description("Enable or disable username & password authentication")
-    .argument("<enable|disable>", "enable or disable password authentication")
+    .addArgument(
+      new Argument(
+        "<action>",
+        "enable or disable password authentication",
+      ).choices(["enable", "disable"]),
+    )
```
Paveltarno, line 50: very cool

packages/cli/src/core/resources/auth-config/schema.ts
```diff
@@ -75,6 +75,49 @@ export function hasAnyLoginMethod(config: AuthConfig): boolean {
...
+interface SocialProviderSchema {
+  /** AuthConfig boolean field to toggle (e.g., "enableGoogleLogin") */
+  field: keyof AuthConfig;
+  /** Display label (e.g., "Google") */
+  label: string;
+  /** Custom OAuth configuration, if the provider supports it */
+  customOAuth?: CustomOAuthSchema;
+}
+
+export const SOCIAL_PROVIDERS: Record<string, SocialProviderSchema> = {
```
Paveltarno: nit: does it make sense to change the `string` to a closed provider list?

packages/cli/src/cli/commands/auth/social-login.ts
```diff
@@ -0,0 +1,143 @@
...
+    );
+  }
+
+  // Resolve custom OAuth secret if applicable
+  const oauth = providerInfo.customOAuth;
+  const useCustomOAuth = shouldEnable && hasOAuthOptions && oauth != null;
+  let clientSecret: string | undefined;
+
+  if (useCustomOAuth && oauth) {
+    clientSecret = await resolveSecret({
```
Paveltarno: wdyt about adding an `--env-file` option as well which will only look for the secret `BASE44_GOOGLE_OAUTH_CLIENT_SECRET`? I know it's a bit weird but I'm thinking the user would like to just put all their secrets in a `.env` file (same thing we use for `base44 secrets set --env-file`)

packages/cli/src/core/resources/auth-config/schema.ts
```diff
@@ -75,6 +75,49 @@ export function hasAnyLoginMethod(config: AuthConfig): boolean {
...
+  /** AuthConfig field for the OAuth mode (e.g., "googleOAuthMode") */
+  modeField: "googleOAuthMode";
+  /** AuthConfig field for the client ID (e.g., "googleOAuthClientId") */
+  clientIdField: "googleOAuthClientId";
+  /** Secret key name as expected by the backend */
+  secretKey: string;
+  /** Environment variable name for the client secret */
+  envVar: string;
+  /** Prompt message for interactive secret input */
+  promptMessage: string;
```
Paveltarno: these fields should probably not be in `/core` as they are very CLI oriented

packages/cli/src/core/resources/auth-config/schema.ts
```diff
@@ -75,6 +75,49 @@ export function hasAnyLoginMethod(config: AuthConfig): boolean {
...
+  /** Display label (e.g., "Google") */
+  label: string;
+  /** Custom OAuth configuration, if the provider supports it */
+  customOAuth?: CustomOAuthSchema;
+}
+
+export const SOCIAL_PROVIDERS: Record<string, SocialProviderSchema> = {
+  google: {
+    field: "enableGoogleLogin",
+    label: "Google",
```
Paveltarno: doesn't bother me that much but this is also a UI concern, maybe we should have a mapping/transformation for the key?

packages/cli/src/core/resources/auth-config/schema.ts
```diff
@@ -75,6 +75,49 @@ export function hasAnyLoginMethod(config: AuthConfig): boolean {
...
+      envVar: "BASE44_GOOGLE_OAUTH_CLIENT_SECRET",
+      promptMessage: "Enter Google OAuth client secret",
+    },
+  },
+  microsoft: { field: "enableMicrosoftLogin", label: "Microsoft" },
+  facebook: { field: "enableFacebookLogin", label: "Facebook" },
+  apple: { field: "enableAppleLogin", label: "Apple" },
+};
+
+export const VALID_PROVIDER_NAMES = Object.keys(SOCIAL_PROVIDERS);
```
Paveltarno: i think also should move to CLI

packages/cli/src/cli/commands/auth/social-login.ts
```diff
@@ -0,0 +1,143 @@
...
+    )
+    .option("--client-id <id>", "custom OAuth client ID (Google only)")
+    .option(
+      "--client-secret <secret>",
+      "custom OAuth client secret (Google only)",
+    )
+    .option(
+      "--client-secret-stdin",
+      "read client secret from stdin (Google only)",
+    )
```
Paveltarno, line 193: does it make sense to allow setting only the secret without the clientId? afaiu it will just push a redundant secret?

packages/cli/src/cli/commands/auth/social-login.ts
```diff
@@ -0,0 +1,143 @@
...
+      ]),
+    )
+    .option("--client-id <id>", "custom OAuth client ID (Google only)")
+    .option(
+      "--client-secret <secret>",
+      "custom OAuth client secret (Google only)",
+    )
+    .option(
+      "--client-secret-stdin",
+      "read client secret from stdin (Google only)",
```
Paveltarno, line 192: do you think we should also allow the user to set the client-id only and then let them know that they should push the needed secrets using the `base44 secrets set --env-file`? again, im thinking of a scenario where i just have an `.env` file i got from a colleague/vault and i just want to push it once and forget about it

Paveltarno (changes requested):
great! added some comments

packages/cli/src/core/resources/auth-config/schema.ts
moranshe-max: you mean like capitalize utility that transforms the key to the label?

moranshe-max: Added mapping, I prefer the transformation to be explicit.

packages/cli/src/cli/commands/auth/social-login.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  const newStatus = shouldEnable ? "enabled" : "disabled";
+  const oauthNote = useCustomOAuth ? " with custom OAuth" : "";
+
+  let outroMessage = `${label} login ${newStatus}${oauthNote} in local config. Run \`base44 auth push\` or \`base44 deploy\` to apply.`;
+
+  // Hint about pushing secrets separately when client-id was set without a secret
+  if (useCustomOAuth && !clientSecret) {
+    outroMessage += `\nRemember to push the client secret separately: base44 secrets set --env-file <path>`;
```
Paveltarno, line 163: This hint looks wrong as written. `auth social-login ... --env-file` reads `BASE44_GOOGLE_OAUTH_CLIENT_SECRET` and translates it to the backend secret key `google_oauth_client_secret`, but `base44 secrets set --env-file <path>` does not do that translation; it uploads keys verbatim from the file. So if the user follows this hint with a `.env` containing `BASE44_GOOGLE_OAUTH_CLIENT_SECRET=...`, they will store the wrong remote secret name and custom Google OAuth still will not be configured. I think the follow-up should either point back to `auth social-login ... --env-file`, or tell the user to set `google_oauth_client_secret` explicitly.

Paveltarno (changes requested):
i think i found a small issue, not sure, maybe it's worth adding a test to the explicitly

all other stuff looks great!

packages/cli/src/cli/commands/auth/social-login.ts
moranshe-max: Good catch — fixed. The hint now points to `base44 auth social-login <provider> enable --client-id <id> --env-file <path>` which handles the key translation to `google_oauth_client_secret` on the backend.

packages/cli/src/cli/commands/auth/social-login.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  const newStatus = shouldEnable ? "enabled" : "disabled";
+  const oauthNote = useCustomOAuth ? " with custom OAuth" : "";
+
+  let outroMessage = `${label} login ${newStatus}${oauthNote} in local config. Run \`base44 auth push\` or \`base44 deploy\` to apply.`;
+
+  // Hint about pushing secrets separately when client-id was set without a secret
+  if (useCustomOAuth && !clientSecret) {
+    outroMessage += `\nRemember to push the client secret separately: base44 auth social-login ${provider} enable --client-id ${options.clientId} --env-file <path>`;
```
Paveltarno: Thanks for updating the hint. I think I meant something slightly different in my previous comment. My concern is not only what command we suggest here, but that I do want users to be able to put this secret in a `.env` file and then use `base44 secrets set --env-file ...` as a valid workflow. That still does not work cleanly today, because `auth social-login --env-file` reads `BASE44_GOOGLE_OAUTH_CLIENT_SECRET`, while `secrets set --env-file` uploads keys verbatim and the backend secret name is `google_oauth_client_secret`. So users still cannot reuse the same `.env` shape with `secrets set --env-file` unless they know the internal secret key name. I think the fix I had in mind is to standardize the `.env` file key for this flow on the backend secret name (`google_oauth_client_secret`) and update the command/help/tests accordingly.

moranshe-max: Got it. Misunderstood you.
The reason I added the BASE44 prefix is just to make sure it's not being confused with other env vars since google_oauth_client_secret is a pretty generic name. but maybe it's not a real concern?

Paveltarno (approved):
🔥

PR #471, opened 2026-04-09
Add connector automation support to backend functions

Paveltarno:
> [!NOTE]
> ## Description
>
> Adds Zod schema support for connector (webhook-triggered) automations in the function configuration, including a recursive `trigger_conditions` structure for field-level filtering. The schema uses a forward-compatible operator union so that unknown future operators from the backend still parse without errors. An empty `trigger_conditions` form is rejected (mirroring backend validation), while `null`/`undefined` are accepted to clear conditions.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `ConnectorAutomationSchema` to `function/schema.ts` covering `type: "connector"` automations with `integration_type`, `events`, `resource_id`, and `trigger_conditions` fields
> - Added `TriggerConditionSchema` and recursive `TriggerConditionGroupSchema` (via `z.lazy`) for arbitrarily nested and/or condition trees
> - Added `ConditionOperatorSchema` with all known backend operators enumerated plus a forward-compat `z.string()` fallback
> - Included `ConnectorAutomationSchema` in the `AutomationSchema` union alongside existing scheduled and entity types
> - Added `packages/cli/tests/core/function-schema.spec.ts` with 9 tests covering valid, null/undefined, catch-all, recursive, exists-style, and rejection cases
> - Added `.superset/` to `.gitignore`
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The `events` array intentionally has no `.min(1)` constraint — an empty array means "catch-all" for a connector integration, and per-integration enforcement is handled server-side. The `KnownConditionOperators` list is kept in sync with the backend `ConditionOperator` enum; the `z.string().min(1)` fallback ensures forward compatibility when new operators are added.
>
> ---
> <sub>🤖 Generated by Claude | 2026-04-13 08:06 UTC</sub>

### DIFF
--- a/packages/cli/src/core/resources/function/schema.ts
+++ b/packages/cli/src/core/resources/function/schema.ts
@@ -66,10 +66,36 @@ const EntityAutomationSchema = AutomationBaseSchema.extend({
     .min(1, "At least one event type is required"),
 });
 
+// Known operators — kept in sync with backend ConditionOperator enum.
+// The union with z.string() ensures unknown future operators still parse.
+const KnownConditionOperators = [
+  "equals",
+  "not_equals",
+  "gt",
+  "gte",
+  "lt",
+  "lte",
+  "contains",
+  "not_contains",
+  "starts_with",
+  "ends_with",
+  "in_list",
+  "not_in_list",
+  "exists",
+  "not_exists",
+  "is_empty",
+  "is_not_empty",
+] as const;
+
+const ConditionOperatorSchema = z.union([
+  z.enum(KnownConditionOperators),
+  z.string().min(1),
+]);
+
 // Trigger condition (field-level filter)
 const TriggerConditionSchema = z.object({
   field: z.string().min(1),
-  operator: z.string().min(1),
+  operator: ConditionOperatorSchema,
   value: z.unknown().nullable().optional(),
 });
 
--- a/packages/cli/src/core/resources/function/schema.ts
+++ b/packages/cli/src/core/resources/function/schema.ts
@@ -92,28 +92,13 @@ const TriggerConditionGroupSchema: z.ZodType<TriggerConditionGroup> = z.lazy(
     }),
 );
 
-// The backend also accepts `{}` and `{ conditions: [] }` as "clear conditions"
-// requests, so the schema must allow those forms when parsing local config or
-// remote payloads.
-const EmptyTriggerConditionsSchema = z
-  .object({
-    logic: TriggerLogicSchema.optional(),
-    conditions: z.array(z.unknown()).length(0).optional(),
-  })
-  .passthrough();
-
-const TriggerConditionsSchema = z.union([
-  TriggerConditionGroupSchema,
-  EmptyTriggerConditionsSchema,
-]);
-
 // Connector automation (webhook-triggered)
 const ConnectorAutomationSchema = AutomationBaseSchema.extend({
   type: z.literal("connector"),
   integration_type: IntegrationTypeSchema,
   events: z.array(z.string()),
   resource_id: z.string().nullable().optional(),
-  trigger_conditions: TriggerConditionsSchema.nullable().optional(),
+  trigger_conditions: TriggerConditionGroupSchema.nullable().optional(),
 });
 
 // Union of all automation types
--- a/packages/cli/src/core/resources/function/schema.ts
+++ b/packages/cli/src/core/resources/function/schema.ts
@@ -65,12 +65,34 @@ const EntityAutomationSchema = AutomationBaseSchema.extend({
     .min(1, "At least one event type is required"),
 });
 
+// Trigger condition (field-level filter)
+const TriggerConditionSchema = z.object({
+  field: z.string().min(1),
+  operator: z.string().min(1),
+  value: z.unknown(),
+});
+
+// Condition group wrapping an array of conditions
+const TriggerConditionsSchema = z.object({
+  conditions: z.array(TriggerConditionSchema).min(1),
+});
+
+// Connector automation (webhook-triggered)
+const ConnectorAutomationSchema = AutomationBaseSchema.extend({
+  type: z.literal("connector"),
+  integration_type: z.string().min(1, "Integration type cannot be empty"),
+  events: z.array(z.string()).min(1, "At least one event is required"),
+  resource_id: z.string().nullable().optional(),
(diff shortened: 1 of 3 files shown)


Paveltarno:
@claude review

Paveltarno:
## Review Response

Thanks for the thorough review! Here's what we addressed and what we intentionally left as-is:

### Addressed

- **`events` has no minimum — no comment explaining why**: Added inline comment: `// No .min(1) — empty means catch-all; the server enforces per-integration.` (9f32013)
- **No negative-case tests**: Added 3 rejection tests — missing `integration_type`, empty conditions group, unknown automation type. We also already had rejection tests for empty trigger condition forms (`{}`, `{ conditions: [] }`) from 24aa8b5. Total test count is now 154. (9f32013)

### Intentionally left as-is

- **`.strict()` on `TriggerConditionSchema`**: We intentionally removed `.strict()` from all schemas in this file for consistency — no other schema (scheduled, entity, base) uses it. The backend validates malformed payloads; the CLI stays lenient to avoid rejecting future API fields on inbound responses (pull/list).
- **TS type allows empty `conditions` but schema enforces `.min(1)`**: The manual `TriggerConditionGroup` type is required for `z.lazy()` recursion — TypeScript can't infer recursive types from Zod schemas. The schema is the runtime source of truth; the type discrepancy is cosmetic and doesn't affect behavior.
- **No exported types for connector schemas**: These are purely internal to the function schema module. No callers need them today — we can export when needed.

Paveltarno:
@claude re-review

packages/cli/src/core/resources/function/schema.ts
```diff
@@ -65,12 +66,75 @@ const EntityAutomationSchema = AutomationBaseSchema.extend({
...
+    }),
+);
+
+// Connector automation (webhook-triggered)
+const ConnectorAutomationSchema = AutomationBaseSchema.extend({
+  type: z.literal("connector"),
+  integration_type: IntegrationTypeSchema,
+  // No .min(1) — empty means catch-all; the server enforces per-integration.
+  events: z.array(z.string()),
+  resource_id: z.string().nullable().optional(),
```
kfirstri, line 127: I don't see anything uses this

packages/cli/src/core/resources/function/schema.ts
```diff
@@ -65,12 +66,75 @@ const EntityAutomationSchema = AutomationBaseSchema.extend({
...
+  trigger_conditions: TriggerConditionGroupSchema.nullable().optional(),
```
kfirstri, line 128: camelCases 🤩

packages/cli/src/core/resources/function/schema.ts
```diff
@@ -65,12 +66,75 @@ const EntityAutomationSchema = AutomationBaseSchema.extend({
...
+  "not_contains",
+  "starts_with",
+  "ends_with",
+  "in_list",
+  "not_in_list",
+  "exists",
+  "not_exists",
+  "is_empty",
+  "is_not_empty",
+] as const;
```
kfirstri, line 88: are all of these supported on the backend? Maybe we should be more lean about this?

kfirstri (commented):
looks cool :)

packages/cli/src/core/resources/function/schema.ts
Paveltarno: yeah they are all supported, i've decided to add them after talking with the tech lead responsible on the server as he assured me this data is not bound to change soon, thought it would be a better hint for the LLM but can also drop it, wdyt?

packages/cli/src/core/resources/function/schema.ts
Paveltarno: used in some resources, please see preview docs:
https://github.com/base44-dev/mintlify-docs/pull/1091/changes#diff-6f31d7663328dd924c4188958e9de4d3eb8300c37804c3855097eecbc5ba074bR236

packages/cli/src/core/resources/function/schema.ts
Paveltarno: oh now i remember, it's because the existing backend function schema uses snake_case and didn't want to break.. we're actually in a bit of a mix rn, i suggest we tackle this separately 

```
  ┌─────────────┬────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────┐
  │  Resource   │  Config file   │                                             Case                                             │
  ├─────────────┼────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Auth Config │ config.jsonc   │ camelCase (enableGoogleLogin, ssoProviderName)                                               │
  ├─────────────┼────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Connector   │ <type>.jsonc   │ camelCase (type, scopes — single words only though)                                          │
  ├─────────────┼────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Entity      │ <name>.jsonc   │ camelCase (minLength, maxLength, enumNames) + snake_case in RLS (user_condition, created_by) │
  ├─────────────┼────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Agent       │ agent.jsonc    │ snake_case (tool_configs, entity_name, whatsapp_greeting)                                    │
  ├─────────────┼────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Function    │ function.jsonc │ snake_case (function_args, is_active, schedule_mode, integration_type, etc.)                 │
  └─────────────┴────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────┘
```

packages/cli/src/core/resources/function/schema.ts
kfirstri: Lets keep it. I see now in the code that 
```
const ConditionOperatorSchema = z.union([
  z.enum(KnownConditionOperators),
  z.string().min(1),
]);
```

so we're already lean about it with the z.string.min(1) right? it will accept any string?

kfirstri (approved):
lgtm

packages/cli/src/core/resources/function/schema.ts
Paveltarno: correct (:

Paveltarno: correct (:

PR #481, opened 2026-04-15
Security RLS

artemdemo:
> [!NOTE]
> ## Description
>
> This PR implements Row-Level Security (RLS) and Field-Level Security (FLS) enforcement in the local dev server. Entity CRUD routes now resolve the current user from the JWT `Authorization` header and evaluate per-entity and per-field security rules before allowing reads, writes, updates, and deletes. Unauthorized access returns 403 or masked 404 responses to avoid leaking record existence.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `packages/cli/src/cli/dev/dev-server/db/rls.ts` — pure RLS/FLS evaluation engine supporting equality, `$in`, `$nin`, `$ne`, `$all`, `$or`, `$and`, `$nor`, `user_condition`, and `{{user.*}}` template variables
> - Added `packages/cli/src/cli/dev/dev-server/routes/entities/current-user.ts` — resolves the authenticated user from a Bearer JWT by looking up the `sub` claim in the User collection
> - Updated `entities-router.ts` — all CRUD handlers (GET one, GET list, POST, POST bulk, PUT, DELETE one, DELETE bulk) now enforce RLS row filtering and FLS field filtering using the resolved current user
> - Updated `database.ts` — added `getSchema()` method to expose entity schemas to the router layer
> - Updated `main.ts` — fixed `functionManager.reload`/`stopAll` to be properly awaited; randomised test port allocation to avoid collisions when tests run in parallel
> - Added 5 new integration test suites (`dev-security.spec.ts`, `dev-security-in.spec.ts`, `dev-security-ne.spec.ts`, `dev-security-nin.spec.ts`, `dev-security-nor.spec.ts`) covering all RLS operators and operations
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> RLS checks on read operations return `404` (not `403`) to avoid leaking record existence — consistent with how production Base44 apps behave. The FLS engine strips restricted fields from both read responses and write payloads before they reach the database.
>
> ---
> <sub>🤖 Generated by Claude | 2026-04-28 00:00 UTC | 5011eea</sub>


artemdemo:
@claude review

packages/cli/src/core/auth/config.ts
```diff
@@ -23,8 +23,8 @@ let refreshPromise: Promise<string | null> | null = null;
  */
 export async function readAuth(): Promise<AuthData> {
   try {
-    const parsed = await readJsonFile(getAuthFilePath());
-    const result = AuthDataSchema.safeParse(parsed);
+    const authData = await readJsonFile(getAuthFilePath());
+    const result = AuthDataSchema.safeParse(authData);
```
artemdemo, line 27: Not related to PR.
I just think this is better naming

packages/cli/tests/cli/dev.spec.ts
```diff
@@ -38,13 +39,14 @@ describe("dev command", () => {
...
-);
-`,
+      outdent`
+        Deno.serve((req: Request) =>
+          Response.json({
+            authorization: req.headers.get("authorization"),
+            serviceAuthorization: req.headers.get("base44-service-authorization"),
+          }),
+        );
+      `,
```
artemdemo, line 49: Not related to PR.
Adding `outdent` for better formatting

packages/cli/src/cli/dev/dev-server/function-manager.ts
```diff
@@ -93,16 +93,24 @@ export class FunctionManager {
...
+        this.logger.log(`Stopping function: ${name}`);
+        const exited = new Promise<void>((r) => proc.once("exit", () => r()));
+        if (process.platform === "win32" && proc.pid) {
+          spawn("taskkill", ["/pid", String(proc.pid), "/T", "/F"]);
+        } else {
+          proc.kill();
+        }
+        return exited;
+      }),
+    );
```
artemdemo, line 113: Not related to PR, but this fixes flakines (I hope) on windows.
What was happening:
Test in `tests/cli/dev.spec.ts` was failing - this one: "forwards the service token header from Authorization to local functions"

After investigation with AI we assumed following root cause:

1. `RunLiveHandle.stop()` calls `child.kill("SIGINT")` (CLITestkit.ts:252). Windows doesn't support POSIX signals — Node translates this to abrupt termination, so the CLI's `SIGINT→shutdown()` handler in main.ts:235 never runs and `functionManager.stopAll()` is skipped.
2. The spawned Deno process (function-manager.ts:118) is not in a process group on Windows, so it's orphaned with open handles on `.deno/dep_analysis_cache_v2`.
3. `tmp-promise`'s `unsafeCleanup` then tries to delete `<tempDir>/.deno/...` → `EBUSY`.

Why `.deno` lands in the temp dir: testkit sets `HOME=tempDir` (CLITestkit.ts:76), so Deno defaults `DENO_DIR` to `<tempDir>/.deno`.

Solution - to kill process "the windows way"

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
```diff
@@ -111,25 +162,50 @@ export async function createEntityRoutes(
...
           updated_date: now,
         };
 
-        const inserted = stripInternalFields(
-          await collection.insertAsync(record),
+        const inserted = applyFLS(
+          stripInternalFields(await collection.insertAsync(record)),
+          schema,
+          currentUser,
+          "read",
```
artemdemo, line 224: Here and in all cases below - applying FLS check for "read" because at this point I'm returning data to the user.
So I first `create` - it's one permission, but when I return, I actually `read` and it's a different permission, so I need to verify again.
I think this only makes sense of FLS (field level security), but there is no reason to do RLS for that object, since user actually created it.
I only need to make sure that user is not getting technical fields that are not related to him.
Not that important for local dev though, but just in case

packages/cli/src/cli/dev/dev-server/function-manager.ts
kfirstri: Very cool investigation, Did this came up from tests / you checked this on windows?

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
```diff
@@ -33,17 +39,33 @@ export async function createEntityRoutes(
...
-      await handler(req, res, collection);
+      let currentUser: UserDocument | undefined;
+      try {
+        const auth = req.headers.authorization || "";
+        const { payload } =
+          jwt.decode(auth.replace("Bearer ", ""), { complete: true }) ?? {};
+        currentUser = await db
+          .getCollection(USER_COLLECTION)
+          ?.findOneAsync<UserDocument>({ email: payload?.sub });
+      } catch {}
```
kfirstri: nit: extract user fetching to another function, not really related to "withCollection"

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
```diff
@@ -147,7 +223,7 @@ export async function createEntityRoutes(
   router.post(
     "/:entityName/bulk",
```
kfirstri, line 240: Simplification - should maybe the logic for /:entityName and /:entityName/bulk be the same at this point?
It's same logic maintained twice for 1/N items, i guess the 1 item case can use the same bulk function

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
```diff
@@ -160,19 +236,47 @@ export async function createEntityRoutes(
...
+              },
+              currentUser,
+            )
+          ) {
+            res.status(403).json({ error: "Permission denied" });
+            return;
+          }
+
+          let filteredRecord = db.prepareRecord(entityName, record);
+          if (schema) {
```
kfirstri: I dont understand something - why do we check if "schema" exists before `applyFLS`? schema is always defined AFAIU, maybe you mean `if(schema.fls)` or something?

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
```diff
@@ -46,18 +46,18 @@ export function createUserRouter(db: Database, logger: DevLogger): Router {
         const { payload } =
           jwt.decode(auth.replace("Bearer ", ""), { complete: true }) ?? {};
 
-        const result = await db
+        const currentUser = await db
           .getCollection(USER_COLLECTION)
           ?.findOneAsync({ email: payload?.sub });
 
-        if (!result) {
+        if (!currentUser) {
```
kfirstri: Maybe we can make this `withAuth` be a shared util and will be used in both routers? currently this logic is implemented inside the `withCollection`

packages/cli/src/cli/dev/dev-server/db/rls.ts
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Evaluates a user_condition check against user properties.
+ */
+function evaluateUserCondition(
+  condition: Record<string, unknown>,
+  user: Record<string, unknown>,
+): boolean {
+  for (const [key, expected] of Object.entries(condition)) {
+    const userValue = key.startsWith("data.") ? user[key.slice(5)] : user[key];
```
kfirstri: isn't this what `resolveTemplate` does?

kfirstri (commented):
Went over all files other then rls.ts, need to finish going over it

kfirstri (dismissed):

packages/cli/src/cli/dev/dev-server/function-manager.ts
artemdemo: tests on windows

packages/cli/src/cli/dev/dev-server/db/rls.ts
artemdemo: not exactly, but I see your point. Drying it

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
artemdemo: right

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
artemdemo: leftover, removing

packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts
artemdemo: I though about it, but I don't really see how to achieve it.
`withAuth` and `withCollection` have different intent with regarding to the user data.
`withAuth` is not allowing for the user to be undefined, while `withCollection` allows it.
Therefore `withAuth` is throwing errors related to that, while `withCollection` should not.
So the only "shared logic is following:
```ts
const { payload } =
  jwt.decode(auth.replace("Bearer ", ""), { complete: true }) ?? {};

const currentUser = await db
  .getCollection(USER_COLLECTION)
  ?.findOneAsync({ email: payload?.sub });
```
tbh, it doesn't look like worth drying.

packages/cli/src/cli/dev/dev-server/routes/entities/entities-router.ts
artemdemo: ok

kfirstri (approved):

PR #499, opened 2026-04-28
Fix local dev service role token

### LINKED ISSUE #491: [Bug]: base44.asServiceRole in base44 dev depends on caller Authorization instead of a real service token
## Description
In `base44 dev`, server functions created with `createClientFromRequest(req)` do not receive a distinct service-role credential.

For local function requests, the dev proxy copies the incoming `Authorization` header into `Base44-Service-Authorization` instead of providing a real service token. Because of that, `base44.asServiceRole` behaves differently from hosted Base44 environments:

- unauthenticated requests make `base44.asServiceRole` throw because no service token is present
- authenticated requests make `base44.asServiceRole` reuse the caller's user token rather than a true elevated service token

This makes code that works in hosted Base44 server functions fail or behave differently in local development.

## Steps to Reproduce
1. Create a server function that builds a client with `createClientFromRequest(req)` and uses `base44.asServiceRole`, for example:

```ts
import { createClientFromRequest } from "npm:@base44/sdk";

Deno.serve(async (req) => {
  const base44 = createClientFromRequest(req);
  await base44.asServiceRole.entities.User.list();
  return Response.json({ ok: true });
});
```

2. Run `npx base44 dev`.
3. Invoke the function without an `Authorization` header.
4. Observe that `base44.asServiceRole` throws `Service token is required to use asServiceRole. Please provide a serviceToken when creating the client.`
5. Invoke the same function with an `Authorization: Bearer <user-token>` header.
6. Observe that local dev forwards that same value as `Base44-Service-Authorization`, so `asServiceRole` is driven by the caller token instead of a separate service credential.

## Expected Behavior
`base44.asServiceRole` inside server functions should behave consistently between hosted Base44 and `base44 dev`.

## Actual Behavior
In `base44 dev`:
- no service-role token is present for anonymous requests, so `base44.asServiceRole` throws
- for authenticated requests, `Base44-Service-Authorization` is derived from `Authorization`
- local behavior therefore differs from hosted Base44, where the same server-function code works with `base44.asServiceRole`

## Environment
- OS: macOS 26.3
- Node.js version: v22.17.0
- CLI version: 0.0.50

## Error Logs
```shell
Service token is required to use asServiceRole. Please provide a serviceToken when creating the client.
```

## Additional Context
Relevant current implementation in `base44/cli`:
- `packages/cli/src/cli/dev/dev-server/routes/functions.ts` sets `Base44-Service-Authorization` from the incoming `Authorization` header
- `packages/cli/tests/cli/dev.spec.ts` has a test named `forwards the service token header from Authorization to local functions`

Relevant current behavior in `@base44/sdk`:
- `createClientFromRequest(req)` reads `Authorization` and `Base44-Service-Authorization` separately
- `base44.asServiceRole` throws when no `serviceToken` is present

kaguya3222:
## Summary
- Set `Base44-Service-Authorization` to the fixed local dev credential (`Bearer dev`) for proxied dev function requests.
- Preserve the caller's normal `Authorization` header instead of reusing it as the service credential.
- Update the dev command test to cover anonymous and authenticated function calls.

Fixes #491

### DIFF
--- a/packages/cli/src/cli/dev/dev-server/routes/functions.ts
+++ b/packages/cli/src/cli/dev/dev-server/routes/functions.ts
@@ -4,9 +4,12 @@ import type { Request, Response } from "express";
 import { Router } from "express";
 import { createProxyMiddleware } from "http-proxy-middleware";
 import type { DevLogger } from "@/cli/dev/createDevLogger.js";
+import { createJwtToken } from "@/cli/dev/dev-server/auth-token.js";
 import type { FunctionManager } from "@/cli/dev/dev-server/function-manager.js";
 
-const LOCAL_DEV_SERVICE_AUTHORIZATION_TOKEN = "Bearer dev";
+const LOCAL_DEV_SERVICE_AUTHORIZATION_TOKEN = `Bearer ${createJwtToken(
+  "server@server.com",
+)}`;
 
 export function createFunctionRouter(
   manager: FunctionManager,
--- a/packages/cli/src/cli/dev/dev-server/routes/functions.ts
+++ b/packages/cli/src/cli/dev/dev-server/routes/functions.ts
@@ -19,14 +19,11 @@ export function createFunctionRouter(
     on: {
       proxyReq: (proxyReq, req) => {
         const xAppId = req.headers["x-app-id"];
-        const authorization = req.headers.authorization;
 
         if (xAppId) {
           proxyReq.setHeader("Base44-App-Id", xAppId as string);
         }
-        if (authorization) {
-          proxyReq.setHeader("Base44-Service-Authorization", authorization);
-        }
+        proxyReq.setHeader("Base44-Service-Authorization", "Bearer dev");
         proxyReq.setHeader(
           "Base44-Api-Url",
           `${(req as unknown as Request).protocol}://${req.headers.host}`,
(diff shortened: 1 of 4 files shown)


packages/cli/src/cli/dev/dev-server/routes/functions.ts
```diff
@@ -6,6 +6,8 @@ import { createProxyMiddleware } from "http-proxy-middleware";
 import type { DevLogger } from "@/cli/dev/createDevLogger.js";
 import type { FunctionManager } from "@/cli/dev/dev-server/function-manager.js";
 
+const LOCAL_DEV_SERVICE_AUTHORIZATION_TOKEN = "Bearer dev";
```
artemdemo: Have you tested this approach?

artemdemo: It will throw when function code will try to get list of users like this:
```
await base44.asServiceRole.entities.User.list()
```
this is because of this validation: https://github.com/base44/cli/blob/0c764e5d5b67d24c636fc1e2fa2543b1c8311576/packages/cli/src/cli/dev/dev-server/routes/entities/entities-user-router.ts#L28-L55

PR #504, opened 2026-04-30
fix(functions): non-zero exit code if the deploy fails

raywiis:
I want my CI to fail if deployment fails

If there's failures, then the confusing '--force' behavior or pruning old functions will not happen.


Paveltarno (approved):
LGTM! Added some small fixes

PR #516, opened 2026-05-10
fix(dev-server): inject service role token for unauthenticated function calls

yurynix:
> [!NOTE]
> ## Description
>
> Fixes service-role authorization in the local dev server so `asServiceRole` works the same way it does in production. Previously the dev server only forwarded the callers own `authorization` header to functions, so unauthenticated callers (e.g. public-facing subscribe forms) could not perform privileged service-role operations. The dev server now always mints and injects a service-role JWT, and the RLS/FLS engine recognizes the service principal and bypasses access rules accordingly.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `auth/tokens.ts` centralizing JWT minting: `createJwtToken`, `createServiceAuthorizationHeader`, `SERVICE_ROLE_EMAIL` (`server@server.com`), and `isServiceSubject`.
> - Function proxy (`routes/functions.ts`) now always injects a freshly-minted `Base44-Service-Authorization` service token instead of only forwarding the callers `authorization` header, mirroring production behavior so `asServiceRole` works for unauthenticated callers.
> - `resolveCurrentUser` (`routes/entities/current-user.ts`) resolves the service-role JWT subject to a synthetic admin `SERVICE_USER` with `is_service: true`.
> - RLS/FLS engine (`db/rls.ts`) short-circuits to allow when `user.is_service === true`; FLS now treats only `undefined` rules (not all falsy rules) as "no rule", so explicit `false` correctly denies normal users while service users bypass it.
> - Entity routes (`entities-router.ts`) let service users bypass an explicit `false` delete RLS rule, and align read filtering on `!== undefined`.
> - Refactored `auth-router.ts` to import the shared `createJwtToken` instead of defining its own, removing the duplicated `LOCAL_DEV_SECRET`.
> - Sorted CLI exports in `cli/index.ts` (lint fix).
>
> ## Testing
>
>
> ## Checklist
>
> - [x] My code follows the projects style guidelines
>
> ## Additional Notes
>
> Added `tests/cli/dev-rls.spec.ts` covering service-user RLS/FLS bypass, and expanded `tests/cli/dev.spec.ts` with coverage for service-token injection on unauthenticated function calls. The service-role token is local-only and signed with the dev `LOCAL_DEV_SECRET`.
>
> ---
> <sub>🤖 Generated by Claude | 2026-06-04 08:53 UTC | 3d325cb</sub>

### DIFF
--- a/packages/cli/src/cli/dev/dev-server/routes/functions.ts
+++ b/packages/cli/src/cli/dev/dev-server/routes/functions.ts
@@ -4,6 +4,7 @@ import type { Request, Response } from "express";
 import { Router } from "express";
 import { createProxyMiddleware } from "http-proxy-middleware";
 import type { DevLogger } from "@/cli/dev/createDevLogger.js";
+import { createServiceAuthorizationHeader } from "@/cli/dev/dev-server/auth/tokens.js";
 import type { FunctionManager } from "@/cli/dev/dev-server/function-manager.js";
 
 export function createFunctionRouter(
@@ -19,7 +20,6 @@ export function createFunctionRouter(
     on: {
       proxyReq: (proxyReq, req) => {
         const xAppId = req.headers["x-app-id"];
-        const authorization = req.headers.authorization;
 
         if (xAppId) {
           proxyReq.setHeader("Base44-App-Id", xAppId as string);
@@ -29,7 +29,7 @@ export function createFunctionRouter(
         // unauthenticated callers (e.g. public-facing subscribe forms).
         proxyReq.setHeader(
           "Base44-Service-Authorization",
-          authorization ?? "Bearer base44-dev-service-token",
+          createServiceAuthorizationHeader(),
         );
         proxyReq.setHeader(
           "Base44-Api-Url",
--- a/packages/cli/src/cli/dev/dev-server/routes/functions.ts
+++ b/packages/cli/src/cli/dev/dev-server/routes/functions.ts
@@ -24,9 +24,13 @@ export function createFunctionRouter(
         if (xAppId) {
           proxyReq.setHeader("Base44-App-Id", xAppId as string);
         }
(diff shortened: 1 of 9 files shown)


packages/cli/src/cli/dev/dev-server/routes/functions.ts
```diff
@@ -24,9 +24,13 @@ export function createFunctionRouter(
...
         }
-        if (authorization) {
-          proxyReq.setHeader("Base44-Service-Authorization", authorization);
-        }
+        // In production, Base44 always injects a service role token when forwarding
+        // to functions. Replicate that here so asServiceRole works even for
+        // unauthenticated callers (e.g. public-facing subscribe forms).
+        proxyReq.setHeader(
+          "Base44-Service-Authorization",
+          authorization ?? "Bearer base44-dev-service-token",
```
artemdemo: This is not robust enough even for dev.
`authorization` is user authorization. It means that in case user is logged in locally with newly created user (also locally), then overall permissions will be limited. Which is not expected for server side authorization.

I think approach should be more robust than that.
I think `Base44-Service-Authorization` should be actual JWT, maybe simple solution will be to create token with email like `server@server.com` and then in the code check if email is "server" then allow everything. something like that

artemdemo: Same goes for `Bearer base44-dev-service-token` - this token will fail for code that relies on user being presented in JWT:
https://github.com/base44/cli/blob/bbc07f1c77397ef6d18fc7c4eb284e4466ed7922/packages/cli/src/cli/dev/dev-server/routes/entities/current-user.ts#L27

artemdemo (approved):

PR #517, opened 2026-05-10
feat(dev): write .env.local on first run

yurynix:
> [!NOTE]
> ## Description
>
> The `base44 dev` command now writes a `.env.local` file on first run when one does not already exist. Without it, the frontend's `@base44/vite-plugin` has no app ID and falls back to the production backend instead of the local dev server. The file is populated with the app ID and the local dev server URL using the variable names the plugin expects, and existing `.env.local` files are left untouched.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Add `writeEnvLocalIfMissing()` to `dev.ts`, which writes `.env.local` with `VITE_BASE44_APP_ID` and `VITE_BASE44_BACKEND_URL` when the file is absent, and logs that it was created.
> - Resolve the app ID via `initAppConfig()` and point `VITE_BASE44_BACKEND_URL` at the resolved local dev server URL.
> - Skip writing when a `.env.local` already exists so user-maintained files are never overwritten.
> - Capture the project root from `loadResources` and call the new helper after the dev server starts.
> - Extract a `localServerUrl()` helper and reuse it for the outro message.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> Two new integration tests cover both paths: creating `.env.local` when it is missing, and leaving an existing file untouched.
>
> ---
> <sub>🤖 Generated by Claude | 2026-06-04 10:21 UTC | e294dda</sub>

### DIFF
--- a/packages/cli/src/cli/commands/dev.ts
+++ b/packages/cli/src/cli/commands/dev.ts
@@ -1,10 +1,13 @@
 import process from "node:process";
+import { join } from "node:path";
 import type { Command } from "commander";
 import { createDevServer } from "@/cli/dev/dev-server/main.js";
 import type { CLIContext, RunCommandResult } from "@/cli/types.js";
 import { Base44Command, theme } from "@/cli/utils/index.js";
 import { getDenoWrapperPath } from "@/core/assets.js";
+import { initAppConfig } from "@/core/project/app-config.js";
 import { readProjectConfig } from "@/core/project/config.js";
+import { pathExists, writeFile } from "@/core/utils/fs.js";
 
 interface DevOptions {
   port?: string;
@@ -15,6 +18,8 @@ async function devAction(
   options: DevOptions,
 ): Promise<RunCommandResult> {
   const port = options.port ? Number(options.port) : undefined;
+  let projectRoot: string | undefined;
+
   const { port: resolvedPort } = await createDevServer({
     log,
     port,
@@ -22,9 +27,22 @@ async function devAction(
     denoWrapperPath: getDenoWrapperPath(),
     loadResources: async () => {
       const { functions, entities, project } = await readProjectConfig();
+      projectRoot = project.root;
       return { functions, entities, project };
     },
   });
+
+  if (projectRoot) {
+    const envLocalPath = join(projectRoot, ".env.local");
+    if (!(await pathExists(envLocalPath))) {
+      const { id: appId } = await initAppConfig();
+      await writeFile(
+        envLocalPath,
+        `VITE_BASE44_APP_ID=${appId}\nVITE_BASE44_APP_BASE_URL=http://localhost:${resolvedPort}\n`,
+      );
+      log.info("Created .env.local with app ID and dev server URL");
+    }
+  }
 
   return {
     outroMessage: `Dev server is available at ${theme.colors.links(`http://localhost:${resolvedPort}`)}`,
--- a/packages/cli/src/cli/commands/dev.ts
+++ b/packages/cli/src/cli/commands/dev.ts
@@ -1,5 +1,5 @@
-import process from "node:process";
 import { join } from "node:path";
+import process from "node:process";
 import type { Command } from "commander";
 import { createDevServer } from "@/cli/dev/dev-server/main.js";
 import type { CLIContext, RunCommandResult } from "@/cli/types.js";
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli/commands/dev.ts
```diff
@@ -15,17 +18,32 @@ async function devAction(
...
+    const envLocalPath = join(projectRoot, ".env.local");
+    if (!(await pathExists(envLocalPath))) {
+      const { id: appId } = await initAppConfig();
+      await writeFile(
+        envLocalPath,
+        `VITE_BASE44_APP_ID=${appId}\nVITE_BASE44_APP_BASE_URL=http://localhost:${resolvedPort}\n`,
+      );
+      log.info("Created .env.local with app ID and dev server URL");
+    }
+  }
```
artemdemo, line 66: you're dependent on value that is set in `loadResources()`, imo it means you can just run this code in `loadResources()` and not keep it outside.
Maybe you can wrap it in a standalone function though.

packages/cli/src/cli/commands/dev.ts
```diff
@@ -15,17 +18,32 @@ async function devAction(
...
   });
 
+  if (projectRoot) {
+    const envLocalPath = join(projectRoot, ".env.local");
+    if (!(await pathExists(envLocalPath))) {
+      const { id: appId } = await initAppConfig();
+      await writeFile(
+        envLocalPath,
+        `VITE_BASE44_APP_ID=${appId}\nVITE_BASE44_APP_BASE_URL=http://localhost:${resolvedPort}\n`,
+      );
```
artemdemo: I'm not familiar with the needed of this file and  its content.
Is it used somehow during the `dev`?
Why do you want to create it during the `dev`?

yurynix: My understanding is that it's required for `npm run dev`, the vite server

packages/cli/src/cli/commands/dev.ts
```diff
@@ -15,17 +18,32 @@ async function devAction(
...
     },
   });
 
+  if (projectRoot) {
+    const envLocalPath = join(projectRoot, ".env.local");
+    if (!(await pathExists(envLocalPath))) {
+      const { id: appId } = await initAppConfig();
+      await writeFile(
+        envLocalPath,
+        `VITE_BASE44_APP_ID=${appId}\nVITE_BASE44_APP_BASE_URL=http://localhost:${resolvedPort}\n`,
```
artemdemo: Let's introduce function that provides local address, since we output it in 2 places in this file.

packages/cli/src/cli/commands/dev.ts
artemdemo: ☝️ what about this one?

packages/cli/src/cli/commands/dev.ts
artemdemo: Not sure I understand your answer.
Are you sure it's really required?
Which part is requiring it?
How it worked before?

packages/cli/src/cli/commands/dev.ts
yurynix: Wrapped the post-server logic in a standalone `writeEnvLocalIfMissing(projectRoot, port, log)` function. `projectRoot` still comes from `loadResources()` (the resolved port is only known after the server starts, so the write itself has to happen afterwards), but the body is no longer inlined in `devAction`.

packages/cli/src/cli/commands/dev.ts
yurynix: Done — extracted `localServerUrl(port)` and reuse it for both the `.env.local` contents and the outro message.

packages/cli/src/cli/commands/dev.ts
yurynix: Good catch digging in — it turned out the original code had the wrong variable name. The frontend reads its config via `@base44/vite-plugin`, which consumes `VITE_BASE44_APP_ID` and `VITE_BASE44_BACKEND_URL` (same names the platform injects in `clone_s3_repo.py` / the sandbox provider). The previous revision wrote `VITE_BASE44_APP_BASE_URL`, which nothing reads — so the local-backend redirect silently had no effect and the SDK kept hitting production. Before this PR there was simply no `.env.local` on a fresh clone, so you had to write it by hand. Fixed to write `VITE_BASE44_BACKEND_URL`; PR description updated with the full rationale.

artemdemo (dismissed):

artemdemo (approved):

PR #535, opened 2026-06-09
feat(cli): add `init` command for Stripe Projects handoff

moranshe-max:
## What & why

Adds **`base44 init`** — scaffolds a local project for an **already-existing** Base44 app (e.g. one provisioned by the **Stripe Projects CLI**), using the app id and access token supplied via the environment, without creating a new app or requiring an interactive login. This is the local-dev handoff after `stripe projects add base44_projects`.

## Changes

- **`init` command** (`cli/commands/project/init.ts`, registered in `program.ts`): resolves the app id from `--app-id` or `BASE44_APP_ID`, scaffolds the template against the existing id (no `createProject()` call), defaults path to the current dir.
- **Env-token auth** (`core/auth/config.ts`, `core/clients/base44-client.ts`): `BASE44_ACCESS_TOKEN` is used verbatim as the bearer token, taking precedence over `~/.base44/auth.json` and bypassing OAuth refresh (provider-issued tokens can't refresh via `/oauth/token`). `isLoggedIn()` honors it so `requireAuth` commands run without login.
- **`.env` auto-load + normalization** (`core/utils/env.ts` + `cli/bootstrap-env.ts` as the first import): loads `.env`/`.env.local` before the HTTP clients capture the API URL, and normalizes the Stripe CLI's resource-prefixed vars (e.g. `BASE44_PROJECTS_BASE44_APP_ID`) to the bare `BASE44_*` names (prefix-agnostic; ambiguous matches left unset).
- **Non-destructive scaffolding** (`core/project/template.ts`): `renderTemplate` gains a `skipExisting` mode so `init` never clobbers existing files (e.g. a Stripe-managed `.gitignore`).
- **Shared internals** (`cli/commands/project/scaffold-shared.ts`): `create`'s post-scaffold logic (push/deploy/skills/output + `getTemplateById`) is extracted and reused by both `create` and `init` (`create.ts` shrank ~138 lines).

## Verification

- `typecheck` ✓, `lint` ✓, **520 tests** ✓ (new specs: `init`, `env-token-auth`, core `env`; adds a `givenEnv` testkit helper).
- **Real Stripe e2e** against a live provisioned app: `stripe projects env --pull` → `base44 init` (scaffolded the real appId via prefix-normalization) → `entities push` → `deploy --yes`, all authenticated by the Stripe env token; test artifacts cleaned up afterward. (This e2e is what surfaced the prefix-normalization requirement.)

## ⚠️ Known limitation / follow-up

`entities push` (and `deploy`) is a **full reconcile** — the remote is made to match `base44/entities/` exactly, so entities present remotely but absent locally are **deleted**. `init` currently scaffolds a fresh template and does **not** pull the existing app's resources, so running `push`/`deploy` after `init` against an app that already has data models can delete them (including the default `User` entity). There is also no `entities pull` command today.

Recommended follow-up before this is used for non-empty apps: have `init` mirror the existing app's resources (pull entities — `GET /entity-schemas` already exists — plus functions/agents/connectors) so the first push is non-destructive. Tracking this separately.

### DIFF
--- a/docs/api-patterns.md
+++ b/docs/api-patterns.md
@@ -43,6 +43,33 @@ The `base44Client` automatically handles token refresh:
 2. If expired, refreshes token and saves new tokens
 3. On 401 response, attempts refresh and retries once
 
+## Environment-Supplied Credentials (Stripe Projects Handoff)
+
+When `BASE44_ACCESS_TOKEN` is set in the environment, `base44Client` uses it
+verbatim as the bearer token, taking precedence over `~/.base44/auth/auth.json`
+and **bypassing the OAuth refresh logic** (these tokens are provider-issued, not
+Base44 OAuth tokens, so they cannot be refreshed via `/oauth/token`; on a 401 the
+error simply propagates). `isLoggedIn()` also returns `true` when this var is set,
+so `requireAuth` commands run without an interactive login. See
+`getEnvAccessToken()` in `core/auth/config.js`.
+
+This supports the Stripe Projects CLI handoff: it provisions a Base44 app and
+writes the credentials to `.env` via `stripe projects env --pull`. The CLI loads
+`.env`/`.env.local` from the working directory at startup (`loadProjectEnvFiles()`
+in `core/utils/env.js`, wired through `cli/bootstrap-env.js` as the first import so
+it runs before the HTTP clients capture `getBase44ApiUrl()`). Precedence: ambient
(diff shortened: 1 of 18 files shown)


moranshe-max:
Superseded by a 3-PR stack: #536 (scaffold-shared) → #537 (env-token auth) → init (3/3). Same final diff, split for review.

PR #541, opened 2026-06-11
feat(cli): support projectless app id commands

kfirstri:
> [!NOTE]
> ## Description
>
> Adds a global `--app-id` option (bound to the `BASE44_APP_ID` environment variable) so that app-scoped commands can target a Base44 app without being run from a linked local project directory. App context resolution now follows a clear precedence — `--app-id` flag, then `BASE44_APP_ID`, then the local `base44/.app.jsonc` — enabling "projectless" usage for commands like `exec`, `logs`, and `functions list`. Project-creating commands (`create`, `dev`) explicitly reject an app id, and the legacy `--project-id`/`--projectId` flags on `link`/`eject` are renamed to `--app-id` while remaining hidden for backward compatibility.
>
> ## Related Issue
>
> Closes #546
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Registered a global `--app-id <id>` option on the root program, backed by the `BASE44_APP_ID` env var (new `BASE44_APP_ID_ENV_VAR` const).
> - `initAppContext()` / `ensureAppContext()` now accept an optional `appId`, resolving context from the flag/env when present and only falling back to disk otherwise; emits an actionable "No Base44 app ID found" error with hints when nothing resolves.
> - Added `readExplicitAppId()` helper (`app-id-options.ts`) to read the explicit app id and the legacy `--project-id` value, with validation that the two are mutually exclusive.
> - `link` and `eject`: renamed `--project-id`/`--projectId` to `--app-id` (legacy flags kept hidden via `hideHelp()`), updated messaging from "project" to "app", and enabled `showGlobalOptions` in help.
> - `logs`: lists local project functions when run inside a project, otherwise fetches the deployed function list remotely; clearer empty-state messaging.
> - `exec`: now reads the resolved app id from `CLIContext.app` instead of `getAppContext()`.
> - `create` and `dev`: reject `--app-id`/`BASE44_APP_ID` since they operate on a new or linked local project; `dev` now requires a linked project and threads the resolved app id into `.env.local`.
> - `scaffold`: drops its own `--app-id` option in favor of the global one and accepts the env var via the shared resolver.
> - Updated `docs/commands.md` to document the new resolution precedence and `create`/`scaffold` behavior.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The legacy `--project-id` and `--projectId` flags on `link`/`eject` are retained as hidden aliases for backward compatibility and are marked with TODOs to remove once docs and Base44 CLI skills migrate to `--app-id`.
>
> ---
> <sub>🤖 Generated by Claude | 2026-06-17 08:04 UTC | 978e800</sub>


packages/cli/src/cli/commands/project/link.ts
```diff
@@ -243,33 +255,47 @@ async function link(
...
+          "-p, --project-id <id>",
+          "Project ID to link to an existing project",
+        ).hideHelp(),
+      )
+      .addOption(
+        new CommanderOption(
+          "--projectId <id>",
+          "Project ID to link to an existing project",
+        ).hideHelp(),
+      )
```
kfirstri, line 297: do we need both of these options?

packages/cli/src/cli/commands/project/logs.ts
```diff
@@ -172,27 +178,32 @@ function validateLimit(limit: string | undefined): void {
...
 ): Promise<RunCommandResult> {
   validateLimit(options.limit);
   const specifiedFunctions = parseFunctionNames(options.function);
+  const localProjectRoot = ctx.app?.projectRoot;
 
-  // Always read project functions so we can list them in error messages
-  const allProjectFunctions = await getAllFunctionNames();
+  const availableFunctionNames = localProjectRoot
+    ? await getProjectFunctionNames(localProjectRoot)
+    : await getRemoteFunctionNames();
```
kfirstri, line 190: not sure about this condition,

PR #554, opened 2026-06-29
feat(logs): add --follow to stream logs live

ImriKochWix:
> [!NOTE]
> ## Description
>
> Adds a \`--follow\` (\`-f\`) flag to the \`logs\` command that streams new function logs live, polling every 2 seconds and printing only entries that haven't been shown yet. The flag forces ascending (oldest→newest) order and is mutually exclusive with \`--until\` and \`--order\`, since a live tail has no end and always streams in chronological order.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Add \`-f, --follow\` option to the \`logs\` command to stream new logs as they arrive.
> - Implement \`followLogs()\`, a polling loop (2s interval) that fetches logs, deduplicates already-seen entries, sorts them oldest→newest, and writes each line to stdout (JSON or formatted, respecting \`--json\`).
> - Add \`selectNewEntries()\` with \`FollowState\` (a \`lastTime\` watermark plus a \`boundaryKeys\` set) to track the timestamp boundary and avoid re-emitting entries that share the boundary timestamp across polls.
> - Reject \`--follow\` when combined with \`--until\` or \`--order\` (any value) via \`InvalidInputError\`; force \`order = "asc"\` internally for the tail.
> - Export \`LogEntry\`, \`FollowState\`, and \`selectNewEntries\` so the dedup logic can be unit-tested.
>
> ## Testing
>
>
> ## Checklist
>
> - [ ] I have updated \`docs/\` (AGENTS.md) if I made architectural changes
>
> ## Additional Notes
>
> Unit tests cover the \`selectNewEntries\` dedup logic (first poll, boundary dedup, new entries sharing the boundary timestamp, and no-op polls), and CLI integration tests verify that \`--follow\` is rejected when combined with \`--order\` or \`--until\`.
>
> ---
> <sub>🤖 Generated by Claude | 2026-06-30 08:04 UTC | 01e80f2</sub>

### DIFF
--- a/packages/cli/src/cli/commands/project/logs.ts
+++ b/packages/cli/src/cli/commands/project/logs.ts
@@ -26,12 +26,13 @@ interface LogsOptions {
   limit?: string;
   order?: string;
   env?: LogEnv;
+  follow?: boolean;
 }
 
 /**
  * Unified log entry for display.
  */
-interface LogEntry {
+export interface LogEntry {
   time: string;
   level: string;
   message: string;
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli/commands/project/logs.ts
```diff
@@ -215,6 +281,21 @@ async function logsAction(
     };
   }
 
+  if (options.follow) {
+    if (options.until) {
+      throw new InvalidInputError(
+        "--until cannot be combined with --follow (a stream has no end).",
+      );
+    }
+    options.order = "asc"; // tail reads oldest -> newest
```
Paveltarno, line 295: does `follow` play nicely with all the other avilable options? i.e sort?

Paveltarno (commented):
lgtm, left a question

packages/cli/src/cli/commands/project/logs.ts
ImriKochWix: Good catch. Audited each option against `--follow`:

- `--function`, `--since`, `--level`, `--env`, `--json` — compose fine (`--since` seeds the backlog, `--json` → NDJSON).
- `--until` — already rejected (a stream has no end).
- `--limit` — acts as a per-poll cap.
- `--order` — this was the gap: a live tail must stream oldest→newest, so I was forcing `asc` and silently dropping `--order desc`. Now it errors explicitly, mirroring the `--until` guard. `asc`/unset stay a no-op.

Fixed in 0f8ea69, with tests for both conflicts.

Paveltarno (approved):
LGTM!

PR #560, opened 2026-07-07
feat(auth): add workspace API key support

yoavf:
> [!NOTE]
> ## Description
>
> Adds workspace API key authentication so machine principals (CI, provisioning tools) can deploy without an interactive OAuth login. When `BASE44_API_KEY` is set to a workspace key (`b44k_...`), the shared `base44Client` sends it as the `api_key` header and skips OAuth token refresh, without writing `~/.base44/auth/auth.json`. Auth-dependent surfaces (`whoami`, deploy connector reconciliation, 401 hints) are made key-aware.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Add `getWorkspaceApiKeyFromEnv`, `isWorkspaceApiKey`, and `hasWorkspaceApiKeyAuth` helpers in `core/auth/config.ts`, keyed on the `b44k_` prefix.
> - Send `BASE44_API_KEY` as the `api_key` request header in `base44Client` and skip OAuth token refresh / 401 retry when a workspace key is in use.
> - Short-circuit `ensureAuth` middleware for workspace-key auth and tag the error reporter with a `workspace-api-key` principal (no `auth.json` written).
> - Report the workspace key prefix in `base44 whoami` instead of a logged-in email.
> - Give 401s an actionable workspace-key hint (verify the key) instead of the `base44 login` suggestion; helper inlined in `core/errors.ts` to avoid a circular import.
> - Skip the connector reconcile pass on deploy when no connectors are configured and a workspace key is used (those principals get a 403 on the connectors-list endpoint); OAuth users still reconcile removals.
> - Document the workspace API key flow in `docs/api-patterns.md`.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> New `env-token-auth.spec.ts` cases cover: whoami reporting the key prefix, the `api_key` header on API calls, the workspace-key 401 hint, the skipped connectors-list call on deploy, and correctly ignoring a non-`b44k_` `BASE44_API_KEY` when a real OAuth login exists.
>
> ---
> <sub>🤖 Generated by Claude | 2026-07-12 12:02 UTC | cf66e4d</sub>


packages/cli/src/core/clients/base44-client.ts
```diff
@@ -51,6 +54,10 @@ async function handleUnauthorized(
     return;
   }
 
+  if (hasWorkspaceApiKeyAuth()) {
```
guyofeck, line 57: Correct to skip retry (nothing to refresh), but an invalid/expired b44k_ key now surfaces a bare 401 with no hint. Consider a targeted message that links the 401 to unauthorized \ revoked API key

guyofeck (dismissed):
Minor comment

guyofeck:
@yoavf pushed a small follow-up (70807ee) for the 401 case on workspace keys.

A rejected `BASE44_API_KEY` was falling through to the generic 401 hint ("Try logging in again → `base44 login`"), which isn't actionable for a CI principal with no login flow. `getDefaultHints` now detects the workspace key and returns a targeted hint instead. Added a test in `env-token-auth.spec.ts`.

The workspace-key check is inlined in `errors.ts` rather than imported from `core/auth/config` — importing it created a circular dependency (`auth/config` already depends on `errors` for its error classes) that broke class-init order in the bundle. Feel free to fold it into your own follow-up or adjust the wording.

guyofeck:
#skipreview

tomasmaci:
skipreview is not installed in this org, it surfaced a bug of this comment loop. @guyofeck fyi

packages/cli/src/core/project/deploy.ts
```diff
@@ -79,7 +79,8 @@ export async function deployAll(
   });
   await agentResource.push(agents);
   await authConfigResource.push(authConfig);
-  const { results: connectorResults } = await pushConnectors(connectors);
+  const connectorResults =
+    connectors.length > 0 ? (await pushConnectors(connectors)).results : [];
```
ImriKochWix: **Blocking: this breaks connector removal.**

`pushConnectors([])` was not a needless call — it's the reconcile half of deploy. `syncOAuthConnectors` removes remote connectors that are no longer in local config (`connector/push.ts:83-99`), and `syncStripeConnector(undefined)` removes Stripe when it's installed remotely but absent locally (`stripe.ts:35`). With this change, deleting your *last* connector from config and deploying leaves it connected remotely forever — silently. (Removing one-of-several still works since the list stays non-empty.)

The `deploy.spec.ts` change also locks the regression in: mocking the list endpoint as a 403 makes the test fail if the reconcile call ever comes back.

If the motivation is that workspace API keys get a 403 on the connectors-list endpoint, scope the skip to that case (`hasWorkspaceApiKeyAuth() && connectors.length === 0`) or fix the permission server-side — don't change deploy semantics for OAuth users.

yoavf: Good catch — fixed in `ef883f6`. Scoped the skip to `connectors.length === 0 && hasWorkspaceApiKeyAuth()` so OAuth users still reconcile removals; the skip now only applies to workspace keys (which 403 on the connectors-list endpoint). Also reverted the `deploy.spec.ts` mock back to the empty list and moved the 403 case into a new workspace-key deploy test. Full suite green.

ImriKochWix (approved):

PR #563, opened 2026-07-09
fix(functions): include shared modules reachable via parent-escape imports

guyofeck:
> [!NOTE]
> ## Description
>
> When deploying backend functions, shared modules imported from `base44/shared/` (via `../../shared/...`) previously had to be discovered by walking each functions import graph. This PR uploads the entire `base44/shared/` directory wholesale alongside every function so those modules always bundle server-side, with the Deno bundler tree-shaking anything a given function does not import. It also extends the bundled-file glob to include `.jsonc` files, centralized behind a new `BACKEND_FILE_GLOB` constant.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `BACKEND_FILE_GLOB` (`**/*.{js,ts,json,jsonc}`) to `consts.ts` and used it for both function-directory and shared-directory file collection, adding `.jsonc` to the set of bundled file types.
> - Added `readSharedFiles()` in `function/config.ts` to collect every file under `base44/shared/` (returning an empty list when the directory is absent).
> - Merged shared files into each functions `filePaths` (deduped via `Set`) for both configured functions and zero-config (entry-only) functions.
> - Added fixtures under `tests/fixtures/function-shared-imports/` (greet, farewell, hello-sibling functions plus `shared/response.ts` and `shared/constants.jsonc`).
> - Added tests verifying shared files (including `.jsonc`) are uploaded with every function, that same-dir siblings are still included, and that a shared file resolves to the expected function-relative deploy path (`../../shared/response.ts`).
>
> ## Testing
>
>
> ## Checklist
>
> - [x] My code follows the projects style guidelines
>
> ## Additional Notes
>
> Uploading the whole shared directory is simpler and more robust than import-graph walking; correctness relies on the server-side Deno bundler tree-shaking unused modules out of each functions payload.
>
> ---
> <sub>🤖 Generated by Claude | 2026-07-13 13:28 UTC | e8db973</sub>


packages/cli/src/core/resources/function/config.ts
```diff
@@ -17,6 +17,19 @@ import type {
...
+ * Collect every file under `base44/shared/`. These are uploaded alongside each
+ * function so shared modules imported via `../../shared/...` bundle server-side.
+ * The bundler tree-shakes anything a given function doesn't import.
+ */
+async function readSharedFiles(functionsDir: string): Promise<string[]> {
+  const sharedDir = resolve(functionsDir, "..", "shared");
+  if (!(await pathExists(sharedDir))) {
+    return [];
+  }
+  return globby("**/*.{js,ts,json}", { cwd: sharedDir, absolute: true });
```
Paveltarno: do we use this kind of glob for the functions folder as well? 
i think by minimum we can also add `jsonc`

Paveltarno (dismissed):
looks great, left a comment for your consideration

packages/cli/src/core/resources/function/config.ts
guyofeck: Good catch on both. Yes — the function directory used the exact same `**/*.{js,ts,json}` literal (in two spots), so it was duplicated three times total and free to drift. I've extracted it into a single `BACKEND_FILE_GLOB` const used for both the function dir and `base44/shared/`, and added `jsonc` since base44 uses it for config. Added a `.jsonc` shared fixture + assertion to cover it. (0566fb7)

Paveltarno (approved):
🔥

PR #567, opened 2026-07-14
Add workspace support: create/link into a workspace, `workspace list`, `workspace move`

netanelgilad:
> [!NOTE]
> ## Description
>
> Adds first-class workspace (organization) support to the Base44 CLI. Introduces a new `base44 workspace` command group (`list`, `get`, `move`) and lets `create`/`link` target a specific workspace via `--workspace`/`--org`, defaulting to the caller's personal workspace. Workspace selection is server-authoritative: the CLI passes IDs straight through and surfaces the backend's authorization errors rather than pre-validating roles client-side.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **New `base44 workspace` command group** wired into the program: `list` (with a `--role` filter), `get <workspace-id>`, and `move [workspace-id]` (with `--disconnect-integrations` and `-y/--yes`).
> - **New `core/workspace` SDK module** with `listWorkspaces`, `getWorkspace`, and `moveAppToWorkspace`, plus Zod schemas that transform the server's snake_case payload to camelCase. The first workspace returned is flagged `isPersonal`.
> - **`create` and `link` gain `-w/--workspace <id>`** (with a hidden `--org` alias) and a shared `resolveWorkspaceId` helper: pass an explicit ID through, prompt interactively when the user belongs to more than one workspace, otherwise default to the personal workspace with no extra API call.
> - **`createProject` now accepts an `organizationId`** and sends `organization_id` only when targeting a specific workspace.
> - **New `getApp` API + `AppDetailSchema`** to resolve an app's current workspace for the interactive move flow.
> - **Extracted `toJsonStdout` into a shared `cli/utils/json.ts`** so sandbox and workspace commands share one `--json` serializer; all workspace commands honor the global `--json` flag.
> - Added integration tests for the new commands and flows, plus workspace fixtures/handlers in `TestAPIServer`.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> Move and workspace-targeted creation are intentionally server-authoritative — the CLI does not filter by role but relies on the backend to authorize and return clear block reasons, matching the web builder. `getWorkspace` is backed by `listWorkspaces` since there is no per-workspace GET the CLI principal is authorized for.
>
> ---
> <sub>🤖 Generated by Claude | 2026-07-15 15:07 UTC | a56c290</sub>


packages/cli/tests/core/workspace.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect, it } from "vitest";
+import {
+  canCreateAppsInWorkspace,
+  MoveAppResponseSchema,
+  WorkspaceListResponseSchema,
+} from "@/core/workspace/index.js";
+
+describe("workspace schema", () => {
+  it("transforms the workspaces list response to camelCase", () => {
```
netanelgilad: no real need for this test, not interesting, the tests should test real stuff, integrations, e2e product flows etc.
do we have other tests like this? we can probably remove them.

packages/cli/tests/core/workspace.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    expect(parsed).toEqual({
+      success: true,
+      message: "moved",
+      appId: "app-1",
+      newWorkspaceId: "ws-2",
+    });
+  });
+});
+
+describe("canCreateAppsInWorkspace", () => {
```
netanelgilad: we should cover that in a relevant product facing test, not through the unit, for exmaple showing what the actual error would be when trying with one of these roles etc. (can the other roles even be possible?)

packages/cli/tests/cli/workspace_move.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      success: true,
+      message: "App successfully moved to workspace",
+      app_id: "test-app-id",
+      new_workspace_id: "ws-acme",
+    });
+
+    const result = await t.run("workspace", "move", "ws-acme");
+
+    t.expectResult(result).toSucceed();
+    t.expectResult(result).toContain("Acme Inc");
```
netanelgilad: dont use `toContain` in output assertion, use exact assertion so reading the tests shows exactly what the output is in the scenario the test describes

netanelgilad: for the assertions in the tests you added in this PR

packages/cli/src/cli/commands/project/workspace-select.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (flagWorkspaceId) {
+    const workspaces = await fetchWorkspaces(ctx);
+    const match = workspaces.find((w) => w.id === flagWorkspaceId);
+    if (!match) {
+      throw new InvalidInputError(
+        `Workspace "${flagWorkspaceId}" not found, or you are not a member of it.`,
+        { hints: workspaceHints(workspaces) },
+      );
+    }
+    if (!canCreateAppsInWorkspace(match.userRole)) {
```
netanelgilad: Are we protecting against a flow that the server would protect anyway? why do we need clinet side validation for these permission checks?

packages/cli/src/cli/commands/workspace/list.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Base44Command, theme } from "@/cli/utils/index.js";
+import { canCreateAppsInWorkspace, listWorkspaces } from "@/core/index.js";
+import { toJsonStdout, workspaceTag } from "./shared.js";
+
+interface ListOptions {
+  canCreate?: boolean;
+  role?: string;
+}
+
+function pluralize(n: number): string {
```
netanelgilad: do we already have this in the codenase? do we ahv ea place for shared useful utils?

packages/cli/src/cli/commands/workspace/move.ts
```diff
@@ -0,0 +1,178 @@
...
+  target: string | undefined,
+  workspaces: WorkspaceListEntry[],
+  currentWorkspaceId: string | undefined,
+  isInteractive: boolean,
+): Promise<string> {
+  const eligible = workspaces.filter(
+    (w) => canCreateAppsInWorkspace(w.userRole) && w.id !== currentWorkspaceId,
+  );
+
+  if (target) {
```
netanelgilad: let the server validate it nope?

packages/cli/src/cli/commands/workspace/move.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Resolve the target workspace for a move: validate an explicit id, prompt when
+ * interactive, or fail fast in non-interactive mode without one.
+ */
+async function resolveTargetWorkspace(
+  target: string | undefined,
+  workspaces: WorkspaceListEntry[],
+  currentWorkspaceId: string | undefined,
+  isInteractive: boolean,
+): Promise<string> {
+  const eligible = workspaces.filter(
```
netanelgilad: let the server validate it nope?

packages/cli/src/cli/commands/workspace/move.ts
```diff
@@ -0,0 +1,178 @@
...
+          hints: [
+            { message: "Run 'base44 workspace list' to see your workspaces" },
+          ],
+        },
+      );
+    }
+    if (target === currentWorkspaceId) {
+      throw new InvalidInputError("The app is already in that workspace.");
+    }
+    if (!canCreateAppsInWorkspace(match.userRole)) {
```
netanelgilad: let the server validate it nope?

packages/cli/src/cli/commands/workspace/move.ts
```diff
@@ -0,0 +1,178 @@
...
+      {
+        hints: [
+          { message: "Usage: base44 workspace move <workspace-id>" },
+          { message: "Run 'base44 workspace list' to see your workspaces" },
+        ],
+      },
+    );
+  }
+
+  if (eligible.length === 0) {
```
netanelgilad: I would move the task to list workspace to happen here cause it only need in interactive mode

packages/cli/src/cli/commands/workspace/move.ts
netanelgilad: oh now I see that its for something else

packages/cli/src/cli/commands/workspace/shared.ts
```diff
@@ -0,0 +1,12 @@
+import type { WorkspaceListEntry } from "@/core/index.js";
+
+/** Serialize a value as pretty JSON for stdout (the `--json` contract). */
+export function toJsonStdout(result: unknown): string {
```
netanelgilad: this doesnt seem specific to workspace commadn

packages/cli/src/core/project/api.ts
```diff
@@ -92,6 +104,31 @@ export async function listProjects(): Promise<ProjectsResponse> {
   return result.data;
 }
 
+/**
+ * Fetch a single app's details, including the workspace (`organizationId`) it
+ * currently belongs to. Used to show the source workspace before a move.
```
netanelgilad: remove the Used to part, it's inferred byt he usage of the code and the coment could be outdated

davidsu (approved):

packages/cli/src/cli/commands/workspace/move.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export function getWorkspaceMoveCommand(): Command {
+  return new Base44Command("move")
+    .description("Move the current app to another workspace")
+    .argument("[workspace-id]", "Target workspace (organization) ID")
+    .option(
+      "--disconnect-integrations",
+      "Disconnect the app's OAuth integrations as part of the move",
+    )
+    .option("-y, --yes", "Skip the confirmation prompt")
```
carmelc, line 161: Not sure `-y` is the right approach in this case, isn't it non reversible? -y is usually used for FYI "are you sure"'s I would consider -f in this case.

netanelgilad: you can move back. and I guess -yes is already an established pattern in the CLI to skip confirmations, so seems ok to me

PR #573, opened 2026-07-23
feat(push): confirm before destructive resource push

davidsu:
> [!NOTE]
> ## Description
>
> The `push` commands for agents, entities, and connectors are destructive — they overwrite an app's remote resources with the local copy and delete anything not present locally. This PR adds a confirmation guard before any of those pushes runs. Interactively, the user sees a warning describing exactly what will be replaced/deleted and must confirm; in non-interactive mode (`--json`/CI) a new `-y, --yes` flag is required, so the commands fail fast instead of hanging on a prompt.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added a shared `confirmPush()` helper (`packages/cli/src/cli/utils/confirm-push.ts`) that warns and prompts interactively, requires `--yes` in non-interactive mode (throws `InvalidInputError` otherwise), and returns `false` when the user declines.
> - Exported `confirmPush` from the CLI utils barrel (`utils/index.ts`).
> - Wired the guard into the `agents push`, `entities push`, and `connectors push` commands, each with a resource-specific warning message and a `-y, --yes` "Skip confirmation prompt" option; a declined prompt returns a `Push cancelled` outro.
> - Used future-tense ("This will …") wording in the warnings to match `deploy`'s phrasing.
> - Updated push specs to pass `--yes` where auth/push flows are exercised, and added tests asserting each command fails with "--yes is required in non-interactive mode" when the flag is omitted.
> - Tightened the token-refresh-failure test to assert the real "Unauthorized" reason rather than just a non-zero exit.
>
> ## Testing
>
>
> ## Checklist
>
> - [ ] I have updated \`docs/\` (AGENTS.md) if I made architectural changes
>
> ## Additional Notes
>
> The guard mirrors the existing `deploy` command's `--yes` behavior for consistency across destructive commands. Non-interactive callers (CI, `--json`) must now pass `--yes`; this is the intended safety behavior rather than a regression.
>
> ---
> <sub>🤖 Generated by Claude | 2026-07-23 09:20 UTC | 99fe0bb</sub>

### DIFF
--- a/packages/cli/src/cli/commands/agents/push.ts
+++ b/packages/cli/src/cli/commands/agents/push.ts
@@ -60,6 +60,6 @@ export function getAgentsPushCommand(): Command {
     .description(
       "Push local agents to Base44 (replaces all remote agent configs)",
     )
-    .option("-y, --yes", "Skip the confirmation prompt")
+    .option("-y, --yes", "Skip confirmation prompt")
     .action(pushAgentsAction);
 }
--- a/packages/cli/src/cli/commands/agents/push.ts
+++ b/packages/cli/src/cli/commands/agents/push.ts
@@ -25,7 +25,7 @@ async function pushAgentsAction(
     yes: options.yes,
     log,
     warning:
-      "This replaces all remote agent configs with your local agents. Remote agents not present locally will be deleted.",
+      "This will replace all remote agent configs with your local agents and delete any not present locally.",
   });
   if (!proceed) {
     return { outroMessage: "Push cancelled" };
--- a/packages/cli/src/cli/commands/agents/push.ts
+++ b/packages/cli/src/cli/commands/agents/push.ts
@@ -1,13 +1,17 @@
 import type { Command } from "commander";
 import type { CLIContext, RunCommandResult } from "@/cli/types.js";
-import { Base44Command } from "@/cli/utils/index.js";
+import { Base44Command, confirmPush } from "@/cli/utils/index.js";
 import { readProjectConfig } from "@/core/index.js";
 import { pushAgents } from "@/core/resources/agent/index.js";
 
-async function pushAgentsAction({
-  log,
-  runTask,
-}: CLIContext): Promise<RunCommandResult> {
+interface PushOptions {
+  yes?: boolean;
+}
+
+async function pushAgentsAction(
+  { isNonInteractive, log, runTask }: CLIContext,
+  options: PushOptions,
+): Promise<RunCommandResult> {
   const { agents } = await readProjectConfig();
 
   log.info(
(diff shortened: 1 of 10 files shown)


davidsu:
Reviewed (agent `exec-flags`): **approved**. Verified against the agreed design — `confirmPush` mirrors `deploy`'s guard exactly (incl. `InvalidInputError` in non-interactive mode; `isNonInteractive` includes `--json`, so JSON/CI can never hang on the prompt), all three destructive pushes are wired with `-y/--yes`, the connectors warning names remote removal, the entities confirm sits after the zero-entities early return, and each command has a non-interactive-without-`--yes` failure test.

Non-blocking nits: flag help text "Skip the confirmation prompt" vs deploy's "Skip confirmation prompt" — worth aligning; and a follow-up (not this PR) could refactor `project/deploy.ts` to reuse `confirmPush`.

packages/cli/src/cli/commands/agents/push.ts
```diff
@@ -16,6 +20,17 @@ async function pushAgentsAction({
       : `Found ${agents.length} agents to push`,
   );
 
+  const proceed = await confirmPush({
+    isNonInteractive,
+    yes: options.yes,
```
guyofeck, line 25: shouldn't it be `force`?

guyofeck (commented):
I think the convention should be `--force` and not `--yes`, but it's kind of a product decision.
wdyt guys? @davidsu @netanelgilad

packages/cli/src/cli/commands/agents/push.ts
davidsu: I think it should be --yes, here's why:

Existing --yes — 5 commands, all identical label "Skip confirmation prompt": project deploy, site deploy, auth push, workspace move, eject.
Existing --force — exactly 1 command, functions deploy: label "Delete remote functions not found locally"

guyofeck (approved):
--yes is the right flag (not --force). It matches the established convention: project deploy, site deploy, auth push, project eject, and workspace move all expose -y, --yes with the same "Skip confirmation prompt" label, and this PR throws the identical "--yes is required in non-interactive mode" message they do. --force exists on exactly one command (functions deploy) where it means something behaviorally different — "Delete remote functions not found locally" — so reusing it here would be semantically confusing. Keep --yes.

Dropping the non-main-branch gating is correct. base44 create doesn't git init and a project isn't required to be a git repo, so there's no reliable branch signal to key confirmation off of. Gating on branch would be fragile and surprising.

Env var to make --yes permanent (à la dangerously-skip-permissions) is a reasonable ask but out of scope for this PR — fine to track separately. If added, it should be global across all the --yes commands, not push-specific.

Non-blocking follow-up: nice that confirmPush() is a shared helper. The pre-existing destructive commands (project deploy, site deploy, auth push) still duplicate their own inline --yes/throw logic — worth migrating them onto this helper in a later PR so there's one source of truth for the destructive-push confirmation pattern.

PR #576, opened 2026-07-27
feat(dev): run backend functions written against base44:runtime

guyofeck:
> [!NOTE]
> ## Description
>
> Backend functions deploy to Cloudflare workerd and are authored with an `export default` handler plus `import { secrets, waitUntil } from "base44:runtime"`. Deploy already accepted that shape (source ships verbatim, never parsed), but `base44 dev` could not run it: the local wrapper detected readiness only from a patched `Deno.serve`, so a default-export function hung until the readiness timeout, and `base44:runtime` had no local resolver. This PR runs functions on the runtime they actually deploy to — each is bundled into one workerd ESM module (esbuild + `@deno/loader`) and served in workerd via Miniflare, with `base44:runtime` as a virtual module: `secrets` reads real Worker env bindings, `waitUntil` rides `ctx.waitUntil`. The Deno subprocess wrapper remains a fallback runtime, and existing `Deno.serve`/`Deno.env` functions keep working on both.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [x] Documentation update
>
> ## Changes Made
>
> **workerd runtime (new default)**
>
> - `function-bundler.ts` — bundles a function entry into one workerd-ready ESM module. `npm:`/`jsr:`/`https:`/relative specifiers resolve through `@deno/loader`, so versions, multi-version graphs and `exports` maps behave as under Deno; esbuild flattens the graph. `base44:runtime` is virtual, `node:`/`cloudflare:` stay external, unresolvable *optional* deps stay external (so axios-like packages bundle), and a CJS `require()` of a node builtin becomes a static re-export — matching the deployed bundler.
> - Generated Worker entry — installs the `globalThis.Base44` bridge (secrets from the request env binding via `AsyncLocalStorage`, `waitUntil` → `ctx.waitUntil`) and a minimal `globalThis.Deno` (`serve` capture + `env`) before user code runs, then imports the user module lazily inside the first request's store so module-scope secret reads see a live env. A captured `Deno.serve` wins over a default export — the deployed precedence.
> - `miniflare-function-manager.ts` — serves each bundle in workerd on its own port, leaving the proxy in `routes/functions.ts` unchanged. Secrets arrive as Worker env bindings from the environment `base44 dev` started with, so production secrets are never copied locally. Miniflare's SIGINT/SIGTERM handlers `process.exit(128 + signal)` and race graceful shutdown, so those listeners are stripped per instance (its orphan-reaping `exit` hook is kept) and esbuild is stopped on `stopAll` so the event loop drains.
> - `function-runtime.ts` — a `FunctionRuntime` interface plus a factory that probes `import("miniflare")` and falls back to Deno when it cannot resolve, or when forced with `B44_DEV_FUNCTIONS_RUNTIME=deno` (`=workerd` makes failure hard). `base-function-manager.ts` holds the shared lifecycle (name map, start dedupe, ports, teardown).
>
> **Deno fallback runtime**
>
> - Serves `export default` handlers: `backend-runtime/main.ts` resolves a handler off the module namespace (function or `{ fetch }`) and serves it through the same patched `Deno.serve`, so readiness still prints once. A module calling `Deno.serve` on import still wins; one with neither now errors clearly instead of hanging until the timeout.
> - New `backend-runtime/base44-runtime.ts` implements `secrets.get()` and `waitUntil()` with the deployed signatures as a pure delegator over `globalThis.Base44` (installed before the dynamic import; importing with no bridge throws). `function-manager.ts` passes `--import-map` at `backend-runtime/import-map.json` — only that map, not merged with a project `deno.json`, since Deno resolves config from the entry point and deploy uploads only `base44/`.
>
> **Supporting changes**
>
> - Distribution: `miniflare`, `esbuild`, `@deno/loader` ship native binaries/WASM and cannot be bundled, so they become the CLI's only real `dependencies` — external in both builds, imported dynamically. All three are pinned exactly: floating `miniflare@^4` let a fresh install resolve an older 4.x whose workerd rejects the hardcoded `COMPATIBILITY_DATE`, killing every function with "The Workers runtime failed to start" (invisible here, where the lockfile pins newer). `@deno/loader` uses its JSR tarball URL so `npm install base44` works without an `@jsr` `.npmrc`. Binaries exclude all three and fall back to Deno.
> - `deno-runtime/` → `backend-runtime/` (it emulates Cloudflare Workers, not Deno), updating `build.ts`, `build-binaries.ts`, `assets.ts`, `tsconfig.json` and `binary-entry.ts`.
> - Stale-asset refresh: `ensureNpmAssets` and `binary-entry.ts` re-copy when an expected asset directory is missing, not only when the version directory is — otherwise an install on this version keeps the old layout and the wrapper vanishes after the rename.
> - Docs: new `packages/cli/backend-runtime/README.md`, new "Entry file contract" in `docs/resources.md`, `docs/AGENTS.md` rules 6/7 amended for the dependency and dynamic-import exceptions, path/keyword fixes in `docs/binary-distribution.md`. Docs teach one authoring shape (`export default` an async handler); `Deno.serve` support stays a code concern.
> - `.gitignore`: ignore `deno.lock`, a local artifact of running Deno against `backend-runtime/` or fixtures.
>
> ## Testing
>
>
> `tests/cli/dev.spec.ts` drives a live dev server against a default-export function importing `base44:runtime` (handler runs, `secrets.get` reads the env, unset secret is `undefined`, `waitUntil` composes), repeats it under `B44_DEV_FUNCTIONS_RUNTIME=deno`, and adds a `legacy-deno-app` suite for all three `Deno.serve` signatures, `Deno.env.get`, the injected service token and relative imports. `tests/core/assets.spec.ts` covers first-run copy, refresh of an older asset layout, and no re-copy/clobber when assets are current. New fixtures: `with-runtime-api-function/`, `legacy-deno-app/`. Also verified out-of-band: `npm install` of the packed tarball into a clean project, cold-running a function pulling `npm:`/`jsr:` deps with an empty `DENO_DIR` (first request ~2s, warm instant). CI green on `a43251e` across all four `test` matrix jobs (ubuntu/windows × npm/binary); `33bbab7` touches only the pin and a comment.
>
> ## Checklist
>
>
> ## Additional Notes
>
> - Two standing rules gain a documented exception (zero-dependency distribution, no dynamic imports), both forced by native binaries/WASM and now written into `docs/AGENTS.md`.
> - `packages/cli/package.json` looks rewritten only because adding `dependencies` reformatted it from 4-space to 2-space indent; that block is the only content change.
> - `COMPATIBILITY_DATE` and the pinned `miniflare` version must be bumped together.
> - `backend-runtime/` is covered by neither Biome nor `typecheck` (Deno types): `deno check --import-map ./import-map.json main.ts base44-runtime.ts exec.ts`.
> - `main.ts`, `import-map.json` and `base44-runtime.ts` must stay co-located — the import map is resolved relative to the wrapper.
> - `base44 exec` still runs on Deno via `exec.ts`; only `base44 dev` functions moved to workerd.
> - Deployed, a few reserved secret names are filtered to `undefined`; not reproduced locally.
> - Projects still activated on `v1`/`v2` have no bundler, so `base44:runtime` will not resolve and a default-export handler is never served there, though both shapes work under `base44 dev`. New apps activate on Cloudflare.
>
> ---
> <sub>🤖 Generated by Claude | 2026-07-28 07:22 UTC | 33bbab7</sub>


guyofeck:
## Validation summary — both conventions, locally and in production

Verified against a real app created with the **published 0.1.5 CLI** (`base44-lab-runtime-check`), then deployed and exercised in production and under `base44 dev`. Test functions deliberately combine `jsr:`, `npm:`, relative imports, `@base44/sdk`, `Deno.env`, and `base44:runtime`.

### Conclusions

1. **A new-syntax function works locally.** `export default` + `base44:runtime` (`secrets`, `waitUntil`) + `jsr:` + `npm:` + relative imports serves correctly under `base44 dev`. The same file on published 0.1.5 fails with `Unsupported scheme "base44"` and exits 1, so this is a genuinely new capability. Locked in by an integration test through the real dev server.
2. **Existing Deno-syntax apps are unaffected.** All three `Deno.serve` signatures, `Deno.env.get`, `jsr:`/`npm:` and relative imports behave identically before and after. `functions deploy` from this branch reports `unchanged` against artefacts uploaded by the published CLI — byte-identical payloads, so the deploy path is provably untouched.
3. **Local matches production for a `cfw` app.** The new-syntax function returns the same fields in both, differing only in the secret's value. Local success predicts deployed behaviour.
4. **Not a breaking change.** Exactly one behaviour changes: a default-export function under `base44 dev` goes from a 30s hang to serving.

### Measured matrix

| | `legacy-deno`<br>`Deno.serve(handler)` | `legacy-serve-object`<br>`Deno.serve({handler})` | `new-runtime`<br>`export default` + `base44:runtime` |
|---|---|---|---|
| `functions deploy` — published 0.1.5 | ✅ | ✅ | ✅ |
| **production** invoke | ✅ | ✅ | ✅ |
| `base44 dev` — published 0.1.5 | ✅ | ✅ | ❌ `Unsupported scheme "base44"` |
| `base44 dev` — **this branch** | ✅ | ✅ | ✅ |
| `functions deploy` — this branch | `unchanged` | `unchanged` | `unchanged` |
| production invoke after branch deploy | ✅ | ✅ | ✅ |

Local run of `new-runtime`:

```json
{"ok":true,"convention":"base44:runtime","shape":"export default","labVersion":"lab-1",
 "jsrPath":"x/y","npmZod":"function","runtimeSecret":"lab-local-value",
 "unsetSecretIsUndefined":true,"composed":"bg-ok","sdkClient":"object"}
```

Two facts confirmed empirically rather than inferred: production `secrets.get` returns `undefined` for an unset name instead of throwing (an earlier revision of this branch threw, inferred from prompt prose about SDK constructors, and was corrected against the bundler source); and `jsr:` specifiers bundle fine on Cloudflare despite the builder prompts teaching `npm:` only.

Also checked by hand: an existing project `deno.lock` is untouched, and a Deno workspace member behaves no worse than baseline.

### Remaining local ↔ production divergences

Neither is introduced by this PR:

1. **Legacy `v1`/`v2` apps.** `base44:runtime` and default-export handlers run locally but have no bundler deployed there. Bounded: `effective_runtime_version` returns `cfw` whenever there is no activated backend project, so this affects already-activated projects only. The CLI cannot detect it — no schema carries the runtime version. Reasoning from apper's source, not measured; no v1/v2 app was available to test.
2. **Project-root `deno.json` aliases.** Fail locally *and* deployed. Deliberately made consistent by dropping an earlier import-map merge (see commit history) rather than supporting locally something that cannot deploy.

### Locally run functions stay on the local dev server

A function's SDK target is decided entirely by the `Base44-Api-Url` header, which `routes/functions.ts` sets to the dev server's own origin; `createClientFromRequest(req)` builds its client from it. Measured with a probe that writes and reads an entity via `asServiceRole`:

| | production | `base44 dev` |
|---|---|---|
| `base44-api-url` seen by the function | `https://…base44.app` | `http://localhost:4400` |
| entity write + list | `["PROD"]` | `["LOCAL1"]` then `["LOCAL1","LOCAL2"]` |

The deployed row never appears locally and local writes accumulate within a dev session (a later session started from an empty store), so the local store is genuinely separate. That probe is itself an `export default` handler, which also confirms the new convention receives the injected service-role token and API URL exactly as a `Deno.serve` function does.

Consequence of this being header-driven (reasoning, not measured): a function building its client with `createClient({ serverUrl: … })` instead of `createClientFromRequest(req)` bypasses the injection and will talk to whatever it hardcodes — production included — even under `base44 dev`.



### `base44:runtime` is a delegator over a host-installed bridge

Follow-up to review feedback: the local module originally read `Deno.env` itself, which worked but diverged structurally from the deployed one. Deployed, `base44:runtime` holds no state and reads no environment — it delegates to a `globalThis.Base44` bridge the generated Worker entry installs. The local module now mirrors that exactly, with `main.ts` as the host:

| | host | secrets source | `waitUntil` |
|---|---|---|---|
| deployed | generated Worker entry | request's Worker env binding (per-request `AsyncLocalStorage`) | `ctx.waitUntil` |
| local | `main.ts` | this process's environment | tracked; rejections logged |

Same shape on both sides, so changing what a secret *is* only touches the host — swapping the local source later needs no change to `base44-runtime.ts`. The bridge is installed before the dynamic import, so a module-scope secret read works; importing with no bridge present throws a clear error rather than failing obscurely.

Verified after the refactor: module-scope secret reads, a `waitUntil` rejection being reported without failing the request, legacy `Deno.serve` unaffected, and the full lab app re-run under `base44 dev` with byte-identical responses.

### Definition of done — `base44:runtime` works locally

| # | Criterion | Status |
|---|---|---|
| 1 | `import { secrets, waitUntil } from "base44:runtime"` resolves under `base44 dev` | ✅ |
| 2 | Both handler conventions serve; `Deno.serve` wins when both present, matching the bundler | ✅ |
| 3 | Signatures match deployed: `secrets.get(name): string \| undefined`, `waitUntil<T>(p): Promise<T>` | ✅ |
| 4 | Legacy `Deno.serve` / `Deno.env` apps unaffected | ✅ |
| 5 | Ships in both distributions (npm tarball + standalone binary) | ✅ |
| 6 | Module is a thin delegator over a host-installed bridge; no environment access inside it | ✅ |
| 7 | Secrets source is the host's responsibility, swappable without touching the module | ✅ |
| 8 | Tests per convention, plus docs stating the intentional differences | ✅ |

Explicitly **out of scope**: running functions in workerd locally. That needs the bundling pipeline, not just a runtime — workerd cannot resolve `npm:`/`jsr:`/TypeScript, so Miniflare alone could not load a function importing `@base44/sdk`. Deploys go through the bundler service, so the CLI's job is the `base44:runtime` surface, which is what this delivers.

guyofeck:
## Miniflare spike — local runtime on real workerd

Proving out moving `base44 dev` from the Deno subprocess to Miniflare, with `base44:runtime` wired to workerd's own primitives instead of emulated.

### It works, both conventions

Bundled each function with esbuild (virtual `base44:runtime` module + Deno-specifier rewriting), then ran the bundle in Miniflare with secrets supplied as **Worker env bindings**:

```
new syntax → {"convention":"base44:runtime","shape":"export default",
              "secret":"lab-miniflare-value","unsetIsUndefined":true,
              "composed":"bg-ok","jsrPath":"x/y","npmZod":"function"}   200
legacy     → {"convention":"deno","shape":"serve(handler)",
              "denoEnv":"lab-miniflare-value","jsrPath":"a/b"}          200
```

So `secrets` becomes a real env binding and `waitUntil` a real `ctx.waitUntil` — no emulation — while legacy `Deno.serve`/`Deno.env` functions keep working through a `globalThis.Deno` shim. The host installs the same `globalThis.Base44` bridge this PR already introduced, so that refactor carries over unchanged.

### Distribution is the hard constraint

`miniflare` statically imports the `workerd` package (a ~114MB native executable per platform, 5 targets). Measured:

| shape | result |
|---|---|
| npm install (node resolves normally) | ✅ works |
| compiled binary, isolated | ❌ `Cannot find package 'workerd'` |
| compiled binary + `MINIFLARE_WORKERD_PATH` | ❌ resolution happens before the var is read |
| compiled binary + `NODE_PATH` → provisioned dir | ✅ works |

`bun build --compile` embeds miniflare's JS but cannot embed the workerd executable, and resolution is CWD/`NODE_PATH`-relative. So the standalone binary needs `workerd` + `@cloudflare/workerd-<platform>` provisioned into a known location (the existing `~/.base44/assets/` pattern fits) with `NODE_PATH` pointed at it. Note this is *not* a PATH prerequisite like Deno — it is a node package that must be resolvable.

### Still to do before this can replace the Deno subprocess

- **Real specifier resolution.** The spike rewrote `npm:`/`jsr:` with a regex, which ignores versions — `npm:zod@3.23.8` resolved to whatever single `zod` was installed. Parity needs `@deno/loader` (available as `npm:@jsr/deno__loader`, requires `@jsr:registry=https://npm.jsr.io`); apper's adapter is ~200 lines including security hardening, optional-dep leniency and error attribution.
- workerd provisioning + `NODE_PATH` plumbing, with integrity checks.
- Gating for `v1`/`v2` apps, which genuinely run Deno in production and would get *worse* local fidelity under workerd — the CLI cannot currently detect an app's runtime.
- Reworking readiness/proxying (`routes/functions.ts`), since Miniflare is in-process and needs no `Listening on` handshake.
- `base44 exec` still needs Deno, so Deno does not go away entirely.
- Durable Object wiring, once apper#13396 lands.

The dependencies install cleanly (`esbuild@0.28.0`, `miniflare@^4`, `@deno/shim-deno@0.19.2`, `@deno/loader@npm:@jsr/deno__loader@^0.5.0`) but must be real `dependencies` rather than bundled devDependencies, which changes the CLI's zero-dependency distribution model.

guyofeck:
## `base44 dev` now runs functions on workerd (Miniflare)

Follow-up to the spike above — implemented, in this PR. Local functions now run on the runtime they deploy to, not an emulation.

### How it works

```
entry.ts ──► @deno/loader (Deno's own resolver: npm:/jsr:/https:, versions,
             exports maps — fetches into the Deno cache, no lifecycle scripts)
         ──► esbuild (single workerd ESM module; base44:runtime as a virtual
             module; CJS require() of node builtins → static re-exports,
             same as the deployed bundler)
         ──► Miniflare (real workerd: secrets = Worker env bindings,
             waitUntil = real ctx.waitUntil, one instance per function,
             same proxy/routes as before)
```

`Deno.serve` capture still wins over a default export (deployed precedence), and a minimal `globalThis.Deno` (serve + env) keeps legacy functions working. Binding values come from the environment `base44 dev` was started with — production secrets are never copied locally.

### Verified

- **651 tests pass** (npm mode), including the full dev suite on workerd; **18/18 in binary mode** on the Deno fallback (verified the fallback log line, not just green tests).
- **Lab app e2e on workerd**: all four functions — `Deno.serve` + `jsr:` + `npm:` + `@base44/sdk` + `Deno.env`, `Deno.serve({handler})`, `export default` + `base44:runtime`, and an SDK entity write through the local dev server — byte-identical responses to the Deno runtime and to production.
- lint, typecheck, knip green; binary sizes unchanged (~71–104 MB).

### Distribution consequences (deliberate, documented in AGENTS.md)

- `miniflare`, `esbuild`, `@deno/loader` are now the CLI's only real npm `dependencies` — native binaries/WASM cannot be bundled. They are external in both builds and imported dynamically, so they load only when `base44 dev` runs functions and every other command's startup is untouched.
- `@deno/loader` is declared by its JSR tarball URL; a scoped-registry alias would break `npm install base44` for anyone without an `@jsr` registry entry in .npmrc.
- The **compiled standalone binary** cannot ship workerd, so it falls back to the Deno wrapper (which still supports both conventions via the import-map shim from earlier in this PR). `B44_DEV_FUNCTIONS_RUNTIME=deno|workerd` forces either runtime.

### Two fights with miniflare worth knowing about

- It registers SIGINT/SIGTERM handlers that `process.exit(130)` immediately, racing the dev server's graceful shutdown. The listeners each instance adds are diffed and stripped (its `exit` hook that reaps workerd orphans is kept).
- The esbuild service child process keeps the Node event loop alive; it's stopped on `stopAll()` so shutdown drains to exit 0 (the CLI never calls `process.exit`).

### What this sets up

Durable Objects: when user-function DO support lands platform-side (apper#13396), the local side needs only DO bindings in the Miniflare options plus the bundler emitting the class exports — the runtime is already the real one.

Remaining known gap, unchanged: legacy `v1`/`v2` apps run real Deno in production, so for them workerd-local is *less* faithful; `B44_DEV_FUNCTIONS_RUNTIME=deno` restores the old behaviour. The CLI still cannot detect an app's runtime version.

ImriKochWix (approved):
🔥

PR #579, opened 2026-07-28
fix(dev): accept prod-minted sessions and keep logout on localhost

davidsu:
> [!NOTE]
> ## Description
>
> Fixes two auth gaps in the `base44 dev` local server that left cli-created apps stuck in a broken login loop. First, a token minted by production (e.g. after a Google OAuth round-trip) was rejected locally because `resolveCurrentUser` only knew about users created through local OTP signup, so `User/me` returned 404 and the app stayed logged out despite holding a valid token. Second, the SDK’s `logout()` navigates to `/api/apps/auth/logout?from_url=<app>`, which the dev server bounced to production — prod rejects foreign `from_url` targets and redirects to its own host, stranding the browser on base44.com.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **`routes/entities/current-user.ts`** — when a valid token carries a subject with no local user record, create one on first sight (via a new `createLocalUser` helper) instead of returning `not_found`. The inserted document mirrors the shape written by the OTP registration path (`id`, `email`, `full_name`, `is_service`, `is_verified`, `disabled`, `role`, `collaborator_role`, timestamps), so subsequent requests resolve the same stable user.
> - **`main.ts`** — handle `GET /api/apps/auth/logout` locally, registered *before* the existing `/api/apps/auth/*` redirect carve-out. A new `isLocalRedirectTarget()` guard only honours `from_url` values on `http(s)://localhost` or `127.0.0.1`; anything else falls through to a plain `Logged out` response rather than becoming an open redirect.
> - **`utils.ts`** — extract `deriveFullNameFromEmail()` and reuse it from both `auth-router.ts` (OTP registration) and the new local-user creation path, removing the duplicated regex.
> - **`tests/cli/dev-auth.spec.ts`** — hoist `serverUrl` to suite scope and add three cases: a prod-minted token yields a working session with a stable id across calls, logout 302s back to a localhost `from_url`, and a foreign origin is not used as a redirect target.
> - **`CHANGELOG.md`** — two entries under Fixed describing both behaviours.
>
> ## Testing
>
>
> ## Checklist
>
> - [x] My code follows the project’s style guidelines
>
> ## Additional Notes
>
> The logout route is intentionally answered locally rather than loosening production’s `from_url` sanitisation, which is a shared security control — local sessions are client-side JWTs the SDK has already cleared, so there is no server-side session to tear down. Note that `resolveCurrentUser` decodes rather than verifies the JWT, so the create-on-first-sight path trusts any well-formed token subject; that is consistent with the existing dev-server behaviour and acceptable for a localhost-only development server, but it means the dev user store can accumulate records for arbitrary subjects.
>
> ---
> <sub>🤖 Generated by Claude | 2026-07-28 13:50 UTC | 8e3e3da</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
 ### Fixed
 
 - `base44 link` now lists editor-created apps; previously only apps created by the CLI could be linked.
+- `base44 dev` accepts sessions from production-minted tokens (e.g. Google OAuth round-trips): a valid
+  token whose subject never registered locally now gets a local user record instead of a 404 on `User/me`
+  that left the app stuck logged out.
+- `base44 dev` handles `/api/apps/auth/logout` locally and redirects back to the app's `from_url`;
+  previously the request bounced to production, which drops localhost redirect targets and stranded the
+  browser on base44.com.
 
 ## [0.0.51] - 2026-04-28
 
--- a/packages/cli/src/cli/dev/dev-server/main.ts
+++ b/packages/cli/src/cli/dev/dev-server/main.ts
@@ -30,6 +30,18 @@ import { WatchBase44 } from "./watcher.js";
 const DEFAULT_PORT = 4400;
 const BASE44_APP_URL = "https://base44.app";
 
+function isLocalRedirectTarget(value: string): boolean {
+  try {
+    const { protocol, hostname } = new URL(value);
+    return (
+      (protocol === "http:" || protocol === "https:") &&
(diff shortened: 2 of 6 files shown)


davidsu:
Closing per David's re-scope: login is out of scope for the local-dev epic; superseded by the CLI-owned flow (#580/#581).

PR #580, opened 2026-07-29
feat(create): scaffold backend-and-client on the vite-plugin client convention

davidsu:
> [!NOTE]
> ## Description
>
> The `backend-and-client` scaffold now uses the same client wiring that editor-created Base44 apps use, so both app species share one convention. `vite.config.js` loads `@base44/vite-plugin` (which supplies the `@` alias and the `/api` dev proxy), the SDK client talks to a same-origin `/api` (`serverUrl: ''`), and the app id is no longer baked into scaffolded source — it arrives at run/build time through `VITE_BASE44_APP_ID`. Scaffolded apps therefore get local entities and functions under `base44 dev`, and work unchanged against the production backend under `base44 dev --remote`.
>
> ## Related Issue
>
> None. Related PRs: #586, #587, #588 — the other three slices of the original unified local-dev rescope, already merged; this PR is the remaining template-convergence slice. The `src/lib/app-params.js` copy tracks base44-dev/apper#18730.
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **`templates/backend-and-client/vite.config.js`** — replaced the manual `resolve.alias` block with the `base44()` Vite plugin (`base44()` + `react()`); the plugin supplies the `@` alias and proxies `/api` to the local dev backend.
> - **`templates/backend-and-client/src/lib/app-params.js`** (new) — reads `appId`, `functionsVersion`, and `appBaseUrl` from `import.meta.env`, delegates the token to the SDK's `getAccessToken()`, and honors `?clear_access_token=true` by clearing the stored token. Verbatim copy of the upstream apper file.
> - **`templates/backend-and-client/src/api/base44Client.js`** (new, static) — `createClient` fed from `appParams`, on same-origin `/api` (`serverUrl: ''`, `requiresAuth: false`).
> - **`templates/backend-and-client/src/api/base44Client.js.ejs`** (deleted) — the app id is no longer templated into scaffolded source at create time.
> - **`templates/backend-and-client/package.json`** — bumped `@base44/sdk` to `^0.8.40` (for the `getAccessToken` export) and added `@base44/vite-plugin` `^1.0.30`.
> - **`cli/commands/project/scaffold-shared.ts`** — the post-scaffold build now runs with `VITE_BASE44_APP_ID` set to the new project id, so the first build carries the app id even though it is no longer in source.
> - **`core/site/deploy.ts`** — both missing/empty output-directory hints now point at `base44 build` (which injects the app id) instead of a bare `npm run build`.
> - **`CHANGELOG.md`** — added a `Changed` entry describing the new template convention.
>
> ## Testing
>
>
> Added `create.spec.ts` → "scaffolds backend-and-client with the editor-app client convention", asserting the scaffolded client uses `serverUrl: ''` and `@/lib/app-params` and does **not** hardcode the app id, that `app-params.js` reads `VITE_BASE44_APP_ID` / `VITE_BASE44_APP_BASE_URL`, that `vite.config.js` references `@base44/vite-plugin`, and that the app id still lands in `base44/.app.jsonc`. `build`, `lint`, `typecheck`, and `knip` are green on CI; the four test-matrix jobs (ubuntu/windows × npm/binary) were still pending when this description was generated, so the "all tests pass" box is left unchecked rather than claimed.
>
> ## Checklist
>
>
> ## Additional Notes
>
> - `src/lib/app-params.js` is a verbatim copy of the upstream apper file and must not diverge from it; if that upstream PR changes in review, re-copy the file here.
> - New scaffolds now depend on `@base44/vite-plugin` for the `@` alias and the `/api` proxy. Existing projects are untouched — only newly scaffolded apps lose the hardcoded-`appId` client.
> - Because the app id is injected rather than baked in, a bare `npm run build` in a scaffolded project produces a bundle without an app id; `base44 build` (or `base44 deploy --build`) is the supported path, which is why the deploy hints were reworded.
> - No `docs/` update: this changes template contents and one env var passed to a build, not CLI architecture — the `CHANGELOG.md` entry covers the user-visible behavior.
>
> ---
> <sub>🤖 Generated by Claude | 2026-08-04 11:52 UTC | 14c640a</sub>

### DIFF
--- a/packages/cli/templates/backend-and-client/src/api/base44Client.js
+++ b/packages/cli/templates/backend-and-client/src/api/base44Client.js
@@ -1,0 +1,13 @@
+import { createClient } from '@base44/sdk';
+import { appParams } from '@/lib/app-params';
+
+const { appId, token, functionsVersion, appBaseUrl } = appParams;
+
+export const base44 = createClient({
+  appId,
+  token,
+  functionsVersion,
+  serverUrl: '',
+  requiresAuth: false,
+  appBaseUrl
+});
--- a/packages/cli/templates/backend-and-client/src/api/base44Client.js
+++ b/packages/cli/templates/backend-and-client/src/api/base44Client.js
@@ -8,6 +8,5 @@ export const base44 = createClient({
   token,
   functionsVersion,
   serverUrl: '',
-  requiresAuth: false,
   appBaseUrl
 });
--- a/packages/cli/templates/backend-and-client/src/lib/app-params.js
+++ b/packages/cli/templates/backend-and-client/src/lib/app-params.js
@@ -1,0 +1,28 @@
+import { getAccessToken } from '@base44/sdk';
+
+const isNode = typeof window === 'undefined';
+
+const isClearAccessTokenRequested = () =>
+	!isNode && new URLSearchParams(window.location.search).get("clear_access_token") === 'true';
+
+const clearStoredAccessToken = () => {
+	window.localStorage.removeItem('base44_access_token');
+	window.localStorage.removeItem('token');
+}
+
+const getAppParams = () => {
+	if (isClearAccessTokenRequested()) {
+		clearStoredAccessToken();
+	}
+	return {
+		appId: import.meta.env.VITE_BASE44_APP_ID,
+		token: getAccessToken(),
+		functionsVersion: import.meta.env.VITE_BASE44_FUNCTIONS_VERSION,
+		appBaseUrl: import.meta.env.VITE_BASE44_APP_BASE_URL,
+	}
+}
+
+
+export const appParams = {
+	...getAppParams()
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - `base44 deploy` (and `base44 site deploy`) can now build first: interactive runs ask, and `--build` / `--no-build` pre-answer the prompt.
 - `base44 dev --remote` serves the frontend against the production backend: it runs `site.serveCommand` with `VITE_BASE44_APP_ID` and `VITE_BASE44_APP_BASE_URL` pointing at the app's own published URL, without starting the local backend. Fails if the app has no published URL.
 
+### Changed
+
+- The `backend-and-client` template now scaffolds the same client convention editor-created apps use: `@base44/vite-plugin` + `src/lib/app-params.js`, with the SDK client on same-origin `/api` (`serverUrl: ''`). Under `base44 dev` the plugin proxies `/api` to the local dev backend, so scaffolded apps get local entities and functions; the app id is injected via `VITE_BASE44_APP_ID` by `base44 dev`, `base44 dev --remote`, and the build/deploy commands instead of being baked into source.
+
 ### Fixed
 
 - `base44 link` now lists editor-created apps; previously only apps created by the CLI could be linked.
--- a/packages/cli/src/cli/commands/project/scaffold-shared.ts
+++ b/packages/cli/src/cli/commands/project/scaffold-shared.ts
@@ -103,7 +103,11 @@ export async function completeProjectSetup(
           await execa({ cwd: resolvedPath, shell: true })`${installCommand}`;
 
           updateMessage("Building project...");
-          await execa({ cwd: resolvedPath, shell: true })`${buildCommand}`;
+          await execa({
+            cwd: resolvedPath,
+            shell: true,
+            env: { VITE_BASE44_APP_ID: projectId },
+          })`${buildCommand}`;
 
           updateMessage("Deploying site...");
           return await deploySite(join(resolvedPath, outputDirectory));
--- a/packages/cli/src/core/site/deploy.ts
+++ b/packages/cli/src/core/site/deploy.ts
@@ -16,7 +16,10 @@ export async function deploySite(
       `Output directory does not exist: ${siteOutputDir}. Make sure to build your project first.`,
       {
         hints: [
-          { message: "Run your build command (e.g., 'npm run build') first" },
+          {
+            message:
+              "Run 'base44 build' first (it injects your app id; a bare 'npm run build' does not)",
+          },
         ],
       },
     );
@@ -29,7 +32,10 @@ export async function deploySite(
       siteOutputDir,
       {
         hints: [
-          { message: "Run your build command (e.g., 'npm run build') first" },
+          {
+            message:
+              "Run 'base44 build' first (it injects your app id; a bare 'npm run build' does not)",
+          },
         ],
       },
     );
--- a/packages/cli/templates/backend-and-client/package.json
+++ b/packages/cli/templates/backend-and-client/package.json
@@ -9,7 +9,8 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@base44/sdk": "^0.8.3",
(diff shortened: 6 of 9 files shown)


chimamark199713-dev:
Cryptoease

packages/cli/templates/backend-and-client/src/lib/app-params.js
```diff
@@ -0,0 +1,28 @@
...
+	window.localStorage.removeItem('token');
+}
+
+const getAppParams = () => {
+	if (isClearAccessTokenRequested()) {
+		clearStoredAccessToken();
+	}
+	return {
+		appId: import.meta.env.VITE_BASE44_APP_ID,
+		token: getAccessToken(),
```
netanelgilad, line 19: dont need this here, the SDK handles it internally, and I think we'd rather keep it there

netanelgilad: same in app params PR in apper

packages/cli/templates/backend-and-client/src/lib/app-params.js
```diff
@@ -0,0 +1,28 @@
...
+		functionsVersion: import.meta.env.VITE_BASE44_FUNCTIONS_VERSION,
```
netanelgilad, line 20: I think this one should be read from query parameter and not env var (is there a code path where we inject this as env var? I think we only ever place it as a query parameter, take a look in apper code)

packages/cli/templates/backend-and-client/src/api/base44Client.js
```diff
@@ -0,0 +1,13 @@
...
+import { appParams } from '@/lib/app-params';
+
+const { appId, token, functionsVersion, appBaseUrl } = appParams;
+
+export const base44 = createClient({
+  appId,
+  token,
+  functionsVersion,
+  serverUrl: '',
+  requiresAuth: false,
```
netanelgilad: I think requiresAuth false is the default so I'd rather remove this (ideally in a breaking version I want tor remove this param entirely, so less usages of it the better).
same in apper PR

netanelgilad (dismissed):

packages/cli/templates/backend-and-client/src/api/base44Client.js
davidsu: Confirmed and done upstream. `requiresAuth` is destructured with a default of `false` in the SDK (`dist/client.js:53`), so passing `requiresAuth: false` explicitly is a no-op.

Removed in apper PR #18730 (`base44-dev/apper`), in both `templates/apps_template` and `templates/game-template` copies of `base44Client.js`, per the locked decision that app-params/base44Client changes land upstream first and the CLI template re-copies verbatim. Also resynced the two security-scanner prompt allowlists, which describe the `createClient({...})` shape and would otherwise drift.

The CLI-side copy will pick this up with the re-copy from 18730's tip.

packages/cli/templates/backend-and-client/src/lib/app-params.js
davidsu: I think it's the other way around
https://github.com/base44-dev/apper/blob/2479b59dd6a85bdadd7a2d6a922b7149563f9c2f/backend/app/user_apps/sandbox/providers/sandbox_provider.py#L1076-L1078 sets the env var. I can't find anyplace setting `functions_version`  queryParam, I don't think there is one.

packages/cli/templates/backend-and-client/src/lib/app-params.js
davidsu: right, though looks like something else breaks:
https://github.com/base44-dev/apper/blob/2479b59dd6a85bdadd7a2d6a922b7149563f9c2f/templates/apps_template/src/lib/AuthContext.jsx#L33-L42 -> seems to bypass the sdk and break if app-params doesn't return the token, am I missing something?

https://github.com/base44-dev/apper/blob/2479b59dd6a85bdadd7a2d6a922b7149563f9c2f/backend/app/user_apps/auth_templates/files/OAuthConsent.jsx#L38 -> same, raw fetch (also L89)

we hit this on the apper side - dropped the storage fallback and in-iframe reloads rendered signed out while sdk calls kept working.

netanelgilad: I see, yeah we need to tackle those, they are very problematic.
for the first one - it does
```
import { createAxiosClient } from '@base44/sdk/dist/utils/axios-client';
```
which is very bad by itself as that path is not formal api and we will break it unknowingly 

this whole call should just be in the SDK itself in the frist place, so the template code doesnt try to read the access token and isnt aware of api paths. so let's introduce a new method for it in the SDK.

(more over I dont know why we still have a call to `public_settings` in the template since we moved to protecting private apps on the server. @roymiloh why do we still call it?)

for the second one - I think we can get the token from an initialized client nope? dont we have a method for that?

guyofeck (approved):
Create a github issue for the app params left over and merge it 🔥

packages/cli/templates/backend-and-client/src/lib/app-params.js
davidsu: @netanelgilad went on vacations. @guyofeck and I decided to merge this as is and leave further improvements for way down the line

packages/cli/templates/backend-and-client/src/lib/app-params.js
davidsu: @netanelgilad went on vacations. @guyofeck and I decided to merge this as is and leave further improvements for way down the line

packages/cli/templates/backend-and-client/src/lib/app-params.js
davidsu: Closing this as superseded — the SDK method you asked for shipped.

- `base44.app.getPublicSettings()` in `@base44/sdk@0.8.45` (base44/javascript-sdk#268)
- both apper app templates now call it; the `@base44/sdk/dist/utils/axios-client` import is gone (base44-dev/apper#22508)

`token: getAccessToken()` stays for now. Two consumers still read `appParams.token` directly: `AuthContext.jsx:30` (a boolean gate) and `OAuthConsent.jsx:38,89` (`Bearer` on two raw `/mcp/*` fetches). Both OAuthConsent sites are guarded, so dropping the field would omit the Authorization header rather than break loudly — cookie-backed sessions keep working, bearer-only ones lose auth silently. Not worth that failure mode for a cosmetic win.

Parked deliberately. The clean version is SDK-side: a public `hasToken()` for the gate, plus real methods for `/mcp/consent-info` and `/mcp/authorize-grant` so the template stops touching the token at all.

packages/cli/templates/backend-and-client/src/lib/app-params.js
davidsu: Closing as-is — `functions_version` is env-var only, which is what this template ships.

Re-verified on apper `main` today:
- `backend/app/user_apps/sandbox/providers/sandbox_provider.py:1098` sets `VITE_BASE44_FUNCTIONS_VERSION` (container-creation secret, covered by `test_sandbox_provider_env.py`)
- nothing anywhere sets a `?functions_version=` query param — every `functions_version=` hit in the repo is a Python kwarg, not a URL

So the query-param rung had no producer, and dropping it also removes an attacker-controllable input (`authReturnTo.js` already strips it for that reason).

PR #584, opened 2026-08-02
feat: full-stack deploy — build a commit's worker from the CLI (cf arm)

netanelgilad:
> [!NOTE]
> ## Description
>
> Adds the worker (`cf`) arm of the deployments API so a full-stack framework build (React Router 7, TanStack Start, Astro 6, vinext — anything built through `@cloudflare/vite-plugin`) can be deployed from the CLI as a Workers deployment addressed by the commit that produced it. The CLI detects the `.wrangler/deploy/config.json` redirect artifact, resolves and validates the generated wrangler config, collects the unbundled worker modules plus the static assets, POSTs asset buckets straight to Cloudflare under the upload-session JWT, then finalizes with the worker modules.
>
> The worker and static arms are **one flow with one `createDeployment()` call**, not two lanes: `deployToDeployments()` checks for a worker itself, and that changes only what it sends (the worker's `config` on create, its modules at finalize). Everything downstream follows the server's answer — `uploadDeploymentAssets()` dispatches on the `asset_uploads` discriminant, `finalizeDeployment()` takes whichever payload completes the deployment. Deploy still only *builds*; nothing here publishes, and re-deploying a commit stays idempotent.
>
> ## Related Issue
>
> None (pairs with the `cf` arm of the backend's discriminated create response)
>
> ## Type of Change
>
> - [x] Documentation update
>
> ## Changes Made
>
> - **One flow** (`core/site/deployment.ts`, replacing `static-site.ts`): `deployToDeployments()` — resolve the worker if there is one → manifest → **one** `createDeployment()` → upload → finalize. `config: worker?.config` is the only difference on create, and its presence is what selects the storage target server-side.
> - **One env gate** (`deploymentsApiEnabled()`, `BASE44_DEPLOYMENTS_API`): the gate is the *only* thing that selects the transport — deployments API for static output and full-stack builds alike, legacy tar.gz otherwise. It is read in exactly two places, both in the command: the transport choice, and the registration of `--git-hash` / `--concurrency`, which exist only on the lane that can honor them. On the lane a worker build needs no `site.outputDirectory`, since the worker brings its own assets directory.
> - **Artifact detection** (`core/site/wrangler-config.ts`): the only trigger is `.wrangler/deploy/config.json`, with `configPath` resolved relative to the redirect file's own directory; hand-authored root configs are deliberately not artifacts. The resolved config must be `no_bundle: true`, and only fields the deploy acts on are declared, so bindings and the worker `name` ride along ignored.
> - **Module collection** (`core/site/modules.ts`): entry module plus the `rules` globs, relative paths as module names, `wrangler.json` / `.dev.vars` / the assets dir excluded, sourcemaps beside modules (all of them under `upload_source_maps`), 40 MB total cap under the server's 50 MB.
> - **Commit address** (`core/site/git-hash.ts`): `resolveGitHash()` — explicit `--git-hash` wins, otherwise `git rev-parse HEAD`; neither available fails fast with guidance. A malformed value is still rejected by the option's `argParser` before the action runs.
> - **`cf` create arm** (`core/site/schema.ts`): `asset_uploads` becomes a discriminated union (`cf` alongside `s3`), the request carries an optional worker `config`, and `AssetUploadResponseSchema` parses Cloudflare's bucket reply.
> - **Direct Cloudflare uploads** (`core/site/upload.ts`): buckets POST straight to Cloudflare with `?base64=true` and a bearer session JWT — never the app client, which would leak app auth. 401/403 maps to "upload session expired — rerun deploy". The completion token is read from whichever bucket reply carries one, since the server decides completeness by manifest membership and buckets upload concurrently. Both arms now share one `UPLOAD_RETRY` on ky (the cf arm must name `methods: ["post"]` or bucket uploads would never retry).
> - **One finalize** (`core/site/api.ts`): `finalizeDeployment(deploymentId, sessionId, payload)` where payload is `{modules, completionJwt}` or `{indexHtml}`; the `if` lives inside the multipart form builder, where the protocol itself differs.
> - **Warnings, not silent drops**: `_headers`/`_redirects` and `run_worker_first` route arrays are surfaced as warnings after the task rather than erroring or quietly changing behavior. `vars` and a missing `nodejs_compat` are passed over untouched — the config is framework-generated, so the fix belongs in the adapter's settings.
> - **Content types** (`core/site/manifest.ts`): per-file MIME lookup for the cf arm's multipart parts; the s3 arm still echoes the server-signed `Content-Type` verbatim.
> - **Command surface** (`cli/commands/site/deploy.ts`): the build step now runs *before* the deploy confirmation, so the prompt is the last gate before anything leaves the machine; `maybeBuildBeforeDeploy` no longer requires `site.outputDirectory`; the flow keeps saying "site" to the user whether or not there is a worker; and the tar.gz path is now the one that raises "No site configuration found."
> - **Docs**: `docs/deployments.md` rewritten to cover the single flow and both arms; `docs/resources.md`, `docs/testing.md`, `docs/AGENTS.md` updated.
>
> ## Testing
>
>
> New: `tests/cli/fullstack_deploy.spec.ts` (manifest/bucket/finalize round trip, null completion token when every asset is already stored, git-hash requirement and `argParser` rejection, `--json`, a config with no `nodejs_compat` forwarded as-is, bucket retry resending the body, session-expired mapping, create failure), `tests/core/site-modules.spec.ts`, `tests/core/site-wrangler-config.spec.ts`, transport cases in `site_deploy.spec.ts`, a `base44 deploy` case asserting the lane's flags stay off the unified command, and a retargeted `static_site_deployments.spec.ts`. Testkit gains `mockAssetUpload` / `mockAssetUploadAfterFailures` / `mockAssetUploadError` and `assetUploadRequests`, plus a `tests/fixtures/fullstack-project/` fixture.
>
> ## Checklist
>
>
> ## Additional Notes
>
> - **Two behavior changes against `main`.** (1) `main` selected the deployments lane by the *presence* of `--git-hash`; here the env gate picks the transport and the commit defaults to `HEAD`, so with the gate on and no commit available the deploy fails asking for `--git-hash` instead of silently falling back to tar.gz — a build with no address could never be published. (2) The build step moved ahead of the confirmation prompt, and no longer needs `site.outputDirectory` to run.
> - **Backwards compatible otherwise.** With `BASE44_DEPLOYMENTS_API` unset, every `site deploy` takes the legacy tar.gz upload unchanged (a full-stack project's worker is then simply not shipped), and `base44 deploy` (unified) is untouched — it still ships the site through `deployAll()`'s tar.gz step and has neither flag. Adopting the lane there needs a commit address the unified deploy has no way to take.
> - The retry test guards two ky behaviors: POST must be named in `retry.methods`, and ky clones a pristine request so a FormData body survives being resent.
> - Current limitations, each an explicit error or warning rather than a silent drop: output that still needs bundling (`no_bundle !== true`) is rejected; wrangler `vars` are ignored in favor of app secrets; `_headers`/`_redirects` and `run_worker_first` route patterns are not forwarded yet.
> - CI on this head: build, lint, typecheck, knip, publish-preview and the gateway check pass; the four test jobs (ubuntu/windows × npm/binary) were still running when this description was generated, so the testing boxes above are left unchecked rather than asserted.
>
> ---
> <sub>🤖 Generated by Claude | 2026-09-01 03:57 UTC | f81051e</sub>


packages/cli/src/cli/commands/site/deploy.ts
```diff
@@ -34,23 +37,29 @@ async function deployAction(
...
+  if (planned.kind === "none") {
     throw new ConfigNotFoundError("No site configuration found.", {
       hints: [
         {
           message:
             'Add \'site.outputDirectory\' to your config.jsonc (e.g., "site": { "outputDirectory": "dist" })',
         },
+        {
+          message:
+            "Full-stack apps ship from their build artifact — run your framework's build first",
```
netanelgilad: there's no need to have full-stack apps in user facing copy.
the copy that is user facing always says "site" as an encompassiong term for what we deploy, whetehr it has the backend worker or not, so there shoudn't be copy changes ideally

packages/cli/src/cli/commands/site/deploy.ts
```diff
@@ -34,23 +37,29 @@ async function deployAction(
...
+        },
       ],
     });
   }
 
   if (!options.yes) {
     const shouldDeploy = await confirm({
-      message: `Deploy site from ${outputDirectory}?`,
+      message:
+        planned.kind === "full-stack"
```
netanelgilad: same here

packages/cli/src/cli/commands/site/deploy.ts
```diff
@@ -60,58 +69,78 @@ async function deployAction(
 
   await maybeBuildBeforeDeploy(ctx, project, options.build);
 
-  const outputDir = resolve(project.root, outputDirectory);
+  // Planned again: the build may have produced the full-stack artifact that
+  // decides which transport applies.
+  const plan = await planAppDeploy(project);
+
+  switch (plan.kind) {
+    case "full-stack":
```
netanelgilad: I dont think we need to code flow to start separating here based on fullstack / static.
inside deploytoDEplymentsApi we should check if there is a user worker to also deploy and decide internally what to add to the createDeployment call, but the bottom line is that we sdhould have a single call to createDeployment in the code, it's jsut the paramters to it that changes based on what we have to deploy

packages/cli/src/cli/commands/site/deploy.ts
netanelgilad: Reverted to main's copy — the second hint is gone, so the error is `No site configuration found.` with the `site.outputDirectory` hint exactly as before this branch.

Same for the rest of the command's copy: description back to "Deploy built site files to Base44 hosting", spinner/success/error back to "Deploying site..." / "Site deployed" / "Site deploy failed" (no orange highlight for one of them either). I also dropped "full-stack" from the errors the wrangler and module paths raise — "No Workers build artifact found", "Base44 only deploys pre-bundled Workers output", "exceeds the 40 MB limit for a Base44 deploy" — since those reach the user too.

The one place a word had to be chosen rather than reverted: the spinner's `Deploying worker (N modules)…` stage, which only appears when there is a worker. Left as "worker" — it names the thing being uploaded rather than a kind of app, and no other stage line changes with the build. Say so if you'd rather it just stayed "Deploying site...".

Recorded the rule in docs/deployments.md so it doesn't drift back.

---

packages/cli/src/cli/commands/site/deploy.ts
netanelgilad: Done — the prompt is `Deploy site from ${outputDirectory}?` again, with no branch on the kind of build.

It does still need a fallback: a project whose build produced a worker has no `site.outputDirectory` to name (the worker's own `assets.directory` supersedes it — the `fullstack-project` fixture has no site config at all), so with nothing to interpolate the prompt is just `Deploy site?`. That's a missing directory name, not a second kind of deploy.

---

packages/cli/src/cli/commands/site/deploy.ts
netanelgilad: Agreed, and done — there is now one flow and one `createDeployment()` call site.

`core/site/full-stack.ts` and `core/site/static-site.ts` are gone, replaced by `core/site/deployment.ts` with a single `deployToDeployments({projectRoot, outputDir, gitHash, ...})`. It checks for a worker itself, and that only changes what it sends:

```ts
const worker = await resolveWorkerBuild(projectRoot, progress);   // null = no worker
const assetsDir = worker ? worker.assetsDir : outputDir;
const assets = assetsDir ? await buildAssetManifest(assetsDir, appId) : NO_ASSETS;

const created = await createDeployment({
  git_hash: gitHash,
  config: worker?.config,          // presence is what selects the storage target
  asset_manifest: assets.manifest,
});
```

Two more forks collapsed on the way, both for the same reason — they were us deciding something the server had already decided:

- **Upload**: `uploadDeploymentAssets(created.assetUploads, assets, …)` dispatches on the `asset_uploads` discriminant and returns the completion token when that arm has one. It also owns the "Found N static assets (M new)" progress, since counting what's owed is per-arm. The two `assetUploads.type !== "cf"/"s3"` guards each flow used to raise are deleted: with one dispatch there is no wrong arm to reject.
- **Finalize**: one `finalizeDeployment(deploymentId, sessionId, payload)`, where payload is `{modules, completionJwt}` or `{indexHtml}`. The remaining `if` is inside `api.ts` building the multipart form — that difference is the wire protocol (a worker completes with its modules, a static build with the index.html sentinel), not a flow choice, and the bytes on both arms are unchanged.

The command is down to one call: `planAppDeploy()` now returns `deployment | tarball | none` (the two deployments kinds merged), so `site deploy` has a 3-arm switch with a single `deployToDeploymentsApi()`, and its `runDeployTask` wrapper is inlined now that there's one caller. A worker build stays ungated and needs no `site.outputDirectory`.

Net −23 lines. `static_site_deployments.spec.ts` already asserted both parameter shapes of that one call — `expect(body).not.toHaveProperty("config")` on the static arm, `body.config.main` on the worker arm — so the contract you described is locked in by tests. Full suite: 732 passed, all 296 deploy + core tests green, typecheck/lint/knip clean.

---

PR #587, opened 2026-08-03
feat(deploy): opt-in build step — --build / --no-build / interactive ask

davidsu:
> [!NOTE]
> ## Description
>
> `base44 deploy` and `base44 site deploy` can now run the site's `buildCommand` before uploading, so a deploy no longer silently ships a stale bundle. Interactive runs ask "Build the site first?", and the new `--build` / `--no-build` flags pre-answer that prompt (`--build` is the only way to opt in non-interactively). The build reuses the same `runSiteBuild` helper as `base44 build`, so `VITE_BASE44_APP_ID` is injected, a failing build aborts the deploy, and an explicit `--build` errors out instead of silently doing nothing when no `site.buildCommand` is configured.
>
> ## Related Issue
>
> None — follow-up to `base44 build` (#586).
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Added `maybeBuildBeforeDeploy()` to `cli/commands/project/site-build.ts` — a single entry point both deploy commands call. It no-ops when no app is linked; an explicit `--build` always delegates to `runSiteBuild` (which throws `ConfigNotFoundError` with a hint when `site.buildCommand` is missing, so the flag is never silently ignored); otherwise it only considers building when `site.outputDirectory` is configured and asks via `shouldAskToBuild()` (no prompt, and no build, when there is no `buildCommand`, when non-interactive, or when the prompt is cancelled).
> - Wired the step into `project/deploy.ts` (after the deploy summary) and `site/deploy.ts` (before the archive/upload task).
> - Registered `--build` / `--no-build` on both commands. Commander folds these into one `build` option, so an absent flag stays `undefined` and keeps the interactive prompt.
> - Both deploy actions now receive the full `CLIContext` instead of destructuring it, since the helper needs `ctx.app`, `ctx.runTask`, and `ctx.isNonInteractive`.
> - `eject.ts` passes `build: false` to its internal auto-deploy so ejecting never prompts or rebuilds.
> - Tests: new `deploy --build` suite in `tests/cli/build.spec.ts` covering `--build`, `--no-build`, flag-absent-in-non-interactive, `site deploy --build`, a failing build aborting the deploy, and `--build` failing on projects with no `buildCommand` / no `site` config.
> - Fixtures: `with-buildable-site` and `with-failing-build` gained an `outputDirectory` plus a `site-output/index.html` so the deploy path is reachable.
> - CHANGELOG entry under Added.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The build decision lives in the CLI layer because it owns the prompt, matching the existing rule that `core/` stays UI-free. Two behaviours worth a reviewer's eye: in non-interactive mode a missing flag means *no* build, so CI pipelines must pass `--build` explicitly to get one; and `--build` is deliberately strict — it fails the deploy when there is nothing to build, rather than falling through to the upload.
>
> ---
> <sub>🤖 Generated by Claude | 2026-08-03 12:23 UTC | 7ed9b75</sub>


packages/cli/src/cli/commands/project/site-build.ts
```diff
@@ -37,3 +38,29 @@ export async function runSiteBuild(
...
+  isNonInteractive: boolean;
+  buildCommand?: string;
+}
+
+export async function shouldBuildBeforeDeploy({
+  build,
+  isNonInteractive,
+  buildCommand,
+}: BuildBeforeDeployChoice): Promise<boolean> {
+  if (!buildCommand) {
```
yardend-wix: a small ux gap, this returns before checking build, so "deploy --build" with no site.buildCommand silently skips the build with no feedback, whereas "base44 build" errors in the same case.

packages/cli/src/cli/commands/project/deploy.ts
```diff
@@ -97,6 +103,21 @@ export async function deployAction(
     log.info(`Deploying:\n${summaryLines.join("\n")}`);
   }
 
+  if (ctx.app && project.site?.outputDirectory) {
```
yardend-wix: This block (gate → shouldBuildBeforeDeploy → runSiteBuild) is duplicated in site/deploy.ts, maybe worth extracting?

packages/cli/src/cli/commands/project/site-build.ts
davidsu: isn't it reasonable that this will b the behavior? cuz `base44 deploy` has work to do when there's no build and the work is clean. `base44 build` without a build command doesn't make any sense so it should be an error.

yardend-wix (dismissed):

packages/cli/src/cli/commands/project/deploy.ts
davidsu: Extracted — `maybeBuildBeforeDeploy(ctx, project, options.build)` in site-build.ts is now the single copy; both call sites are one line, and `shouldBuildBeforeDeploy` became module-private. a99c5b7

packages/cli/src/cli/commands/project/site-build.ts
yardend-wix: as i see it the issue is when the user explicitly asked for deploy --build, which show the intention to build before deploy, anyway its not so critical, your call

packages/cli/src/cli/commands/project/deploy.ts
davidsu: Settled the explicit-flag case in 4cf5174: `--build` is now honored before any site-config gate, so `deploy --build` with no `site.buildCommand` — or no site at all — fails with the existing "No site build command found" error instead of silently no-opping. The silent paths (no flag, `--no-build`, non-interactive) are unchanged, and the interactive ask still only appears when a buildCommand exists.

yardend-wix (approved):

PR #591, opened 2026-08-06
feat(workflows): read-only workflows commands — list and runs

davidsu:
## What

New `base44 workflows` command group — the CLI answer to "did my scheduled work fail, and why":

```
base44 workflows list                  # workflows with status + run summary
base44 workflows runs                  # runs across the app, newest first
base44 workflows runs --status failed  # failures, each with its underlying error
base44 workflows runs --since 1d -n 50
```

Failed runs print the failing task and the underlying error (e.g. the backend function's HTTP 500). `--json` emits the real records (camelCase, Zod-validated).

## How

- New read-only core module `core/resources/workflow/` (schema + api) over `GET /api/apps/{app_id}/workflows` and `/workflows/runs` — workflows are authored in the builder, so no `Resource<T>`/pull/push.
- `cli/commands/workflows/` — `list`, `runs`, and a shared translation of the workflows-not-enabled 403 into an explanation (apps predating the Workflows system).
- Empty results are disambiguated ("this app has no workflows" vs "no runs match your filters") — one extra GET, only on the empty path.
- `run-now`/test runs are always shown, marked `(test)` / `isTestRun` — an agent verifying its own workflow only has test runs, so hiding them would make empty read as healthy.
- `--since` accepts ISO or relative shorthand (`1h`, `2d`) via `normalizeDatetime`, extracted from `logs.ts` into `cli/utils/datetime.ts` and shared by both commands.

## Tests

- 14 new specs (`workflows_list.spec.ts`, `workflows_runs.spec.ts`) + testkit mocks with query-param capture.
- Live-verified against a real app: real overnight scheduled cron runs, including a scheduled failure surfaced by `--status failed`.

## Notes

- Naming: plural `workflows` matching the existing command families (`functions`, `entities`, `connectors`); Guy's phrasing was singular `workflow` — flagged as an open item, rename is a one-liner if ruled.
- Deferred (recorded in suss-tasks): run detail with `steps[]`, `workflows open`, legacy-automations fallback.
- Companion skills-repo change (reference pages for both commands) is prepared on `base44/skills` branch `feat/workflows-runs-reference`, to merge after the CLI release that ships this.

### DIFF
--- a/packages/cli/src/core/resources/workflow/api.ts
+++ b/packages/cli/src/core/resources/workflow/api.ts
@@ -11,12 +11,19 @@ import {
   ListWorkflowsResponseSchema,
 } from "./schema.js";
 
-export async function listWorkflows(): Promise<ListWorkflowsResponse> {
+export async function listWorkflows(
+  limit?: number,
+): Promise<ListWorkflowsResponse> {
   const appClient = getAppClient();
+
+  const searchParams = new URLSearchParams();
+  if (limit !== undefined) {
+    searchParams.set("limit", String(limit));
+  }
 
   let response: KyResponse;
   try {
-    response = await appClient.get("workflows");
+    response = await appClient.get("workflows", { searchParams });
   } catch (error) {
     throw await ApiError.fromHttpError(error, "listing workflows");
   }
--- a/packages/cli/src/core/resources/workflow/api.ts
+++ b/packages/cli/src/core/resources/workflow/api.ts
@@ -1,0 +1,78 @@
+import type { KyResponse } from "ky";
+import { getAppClient } from "@/core/clients/index.js";
+import { ApiError, SchemaValidationError } from "@/core/errors.js";
+import type {
+  ListWorkflowRunsResponse,
+  ListWorkflowsResponse,
+  WorkflowRunFilters,
+} from "./schema.js";
+import {
+  ListWorkflowRunsResponseSchema,
+  ListWorkflowsResponseSchema,
+} from "./schema.js";
+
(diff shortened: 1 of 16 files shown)


packages/cli/src/core/resources/workflow/api.ts
```diff
@@ -0,0 +1,78 @@
...
+  ListWorkflowRunsResponse,
+  ListWorkflowsResponse,
+  WorkflowRunFilters,
+} from "./schema.js";
+import {
+  ListWorkflowRunsResponseSchema,
+  ListWorkflowsResponseSchema,
+} from "./schema.js";
+
+export async function listWorkflows(): Promise<ListWorkflowsResponse> {
```
yardend-wix: listWorkflows sends no limit so the backend's limit=30 default silently caps this,
listWorkflowRuns do exposes --limit, maybe worth add it for list too?

yardend-wix (dismissed):

packages/cli/src/core/resources/workflow/api.ts
davidsu: Good catch — and it's the same silent-truncation class this PR fixes elsewhere, so doubly right. Added in 69089b4: `-n/--limit` (1-200) on `workflows list`, limit parser shared with `runs` via `commands/workflows/limit.ts`, plus specs for the passthrough and the range validation. Kept the default at the backend's 30 and the output format unchanged — no invented "more may exist" messaging, matching how every other list command here reports counts.

yardend-wix (approved):

PR #593, opened 2026-08-10
fix(site): deploy the site without reading the project's resource files

netanelgilad:
Two changes that came out of diagnosing why every sandbox publish through `base44 site deploy --git-hash` was failing in production.

## 1. `site deploy` reads the config file only

`readProjectConfig()` eagerly reads **and strictly validates** every entity, function, agent, agent-skill, connector, and auth file in the project. `site deploy` consumes none of them — it ships the built output and needs `site.outputDirectory` and `root` — so an unrelated file the CLI can't parse failed the entire deploy.

That is what was happening: Base44 builder apps carry `base44/entities/*.jsonc` written by the builder, some of which `EntitySchema` rejects (`rls.create`, `rls.read`, field shapes). All 58 publish failures on Aug 4 died with `SCHEMA_INVALID` before touching a single asset.

`readProjectSettings()` reads and validates the config file alone (including the plugin-project assertion). `readProjectConfig()` is now defined in terms of it, so there's one path to the config; commands that consume resources keep using `readProjectConfig()` unchanged.

## 2. The `--json` error envelope carries `statusCode` and `requestId`

An `ApiError` already knows the status it got back and the `X-Request-ID` that produced it, but `writeJsonError` dropped both. A caller reading only `error` cannot tell a rejected workspace key from a 500, and has nothing to correlate with server-side logs. The sandbox publish consumes this envelope and tags both onto its Datadog span, so those two fields are the difference between a diagnosable failure and a string.

## Testing

- `bun run test tests/cli/static_site_deployments.spec.ts` — 14 passed, including two new cases: a deploy that succeeds with an invalid `base44/entities/Broken.jsonc` in the project, and a 401 from the deployments API surfacing as `{code: "API_ERROR", statusCode: 401}` on stdout.
- `bun run test` — 68/70 files pass. The two failures (`dev.spec.ts`, `exec.spec.ts`) are environmental in this sandbox: no Deno binary, and `@deno/loader` can't install (jsr registry blocked). Both are unrelated to these changes.
- `bunx tsc --noEmit` — clean apart from the pre-existing `function-bundler.ts` errors from that same missing `@deno/loader`.
- `bun run lint` — clean apart from a pre-existing info on `miniflare-function-manager.ts`.

`docs/deployments.md` records the config-only contract and why.

Backend counterpart: base44-dev/apper#19519, which makes these failures visible in Datadog (errored spans + a facetable failure log) instead of showing `status:ok` on a failed publish.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G1E31AZvYBa86zpcYJFD1g)_


packages/cli/src/core/project/config.ts
```diff
@@ -295,3 +306,20 @@ export async function readProjectConfig(
   const reader = new ProjectConfigReader();
   return await reader.readProjectConfig(projectRoot);
 }
+
+/**
+ * Reads and validates the project config file alone — entity, function, agent,
```
netanelgilad: No need to mention the other files it would fall out of sync when new things are added

netanelgilad: Agreed — fixed in `b2133f7`. Now just "none of the project's resource files", which stays true on its own.

I'd enumerated the same list in two other spots, so those went too: the `site deploy` comment and the config-only paragraph in `docs/deployments.md`.

---

guyofeck (approved):

PR #594, opened 2026-08-10
fix(entities): stop rejecting entity shapes production accepts

netanelgilad:
The CLI's `EntitySchema` re-derives the platform's entity contract by hand, and came out **stricter than the platform**. Since `readProjectConfig()` validates every entity file up front, one such file fails `base44 deploy`, `entities push/pull`, `functions deploy` — every command, whether or not it touches entities.

This isn't hypothetical. Over one afternoon of sandbox publishes, **54 of 58 failures** were this schema rejecting files production accepts and evaluates correctly. Roughly a quarter of publishes in that window.

## The gaps, each pinned to its server-side authority

| What the CLI required | What the platform actually allows | Events |
|---|---|---|
| RLS rule is `boolean \| object` | `OPEN_RULE_VALUES = (None, True, {}, "")` — `null` and `""` mean "open" | 24 |
| `user_condition` keys ∈ `role`/`email`/`id`/`data.*`, values must be strings | Neither the attribute names nor their scalar types are constrained | 11 |
| Field operators ∈ `$in $nin $ne $all` | `SUPPORTED_FIELD_OPERATORS` has **12**: `$eq $ne $in $nin $gt $gte $lt $lte $exists $all $size $elemMatch` | 6 |
| Property `type` must be a string | `validate_json_schema` only requires the key to exist, and doesn't recurse into nested `items.properties`. `["string","null"]` is legal JSON Schema | 13 |
| Name `^[a-zA-Z0-9]+$` | `ENTITY_NAME_PATTERN = ^\w+$` — underscores allowed, and the server's own error message invites them | — |

Authorities: `backend/app/user_apps/entities/rls_validation.py` (self-described "single source of truth" for what the engine accepts), `backend/app/json_schema_utils.py`, `backend/app/user_apps/common/entity_name_validation.py`. Each is now referenced from the code so the next person doesn't re-derive it.

The underscore case appears in no failure above — it's a latent one this closes: any entity named `Order_Item` fails the CLI today.

## `base44 dev` moves with it

Widening the types surfaced two places where the local dev server would have diverged from production on the newly-accepted shapes, so they're fixed here rather than left to drift:

- `checkRLS` treated only `undefined` as open, then fell through to the condition evaluator — which denies anonymous users. So `null` / `""` / `{}` would have **blocked** access locally while production treats them as open. It now honours every `OPEN_RULE_VALUE`.
- The record validator took `property.type` as a string. It now resolves a union by matching any member, with `"null"` added as a member type — which is the entire point of `["string","null"]`.

## Not tightened anywhere

The goal is to stop false rejections: nothing that parsed before stops parsing now. Rejections that reflect a real runtime defect stay — `$contains` still fails, because it matches nothing at runtime.

Two things I deliberately left alone, both worth their own change:
- The platform *also* rejects operators inside `user_condition`, `$regex`/`$where`/`$expr`, `_id`, and un-prefixed custom fields. Mirroring those would be new strictness that breaks someone's working `deploy` today, so it isn't in a PR whose thesis is the opposite.
- **Dotted entity names are genuinely invalid** — `commerce.Coupon` / `commerce.Cart` (the remaining 4 of the 58) fail `^\w+$` server-side too. Those two apps have names the platform's own validator would reject, so they either predate it or arrived through a path that skipped `_validate_entity_name_and_schema`. That's data drift to chase separately; the CLI is right to reject them.

## Testing

- `tests/core/entity-schema.spec.ts` — 7 cases, every shape taken from a real production file this schema rejected, plus the two rejections that must survive.
- `tests/core/dev-entity-rules.spec.ts` — 6 cases over `checkRLS` and the validator directly, since `dev.spec.ts` needs a Deno binary that isn't available in this sandbox.
- `bun run test`: 70/72 files, 703 passed. The two failures are `dev.spec.ts` and `exec.spec.ts`, both environmental here (no Deno; `@deno/loader` can't install through the proxy) and failing identically on `main`.
- `bunx tsc --noEmit` clean apart from the pre-existing `function-bundler.ts` errors from that same missing `@deno/loader`; `bun run lint` clean apart from a pre-existing info.

Found while investigating base44-dev/apper#19519, which routed `site deploy` *around* this validator. That unblocked the publish path; this fixes the validator itself, which is what unblocks every other command.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G1E31AZvYBa86zpcYJFD1g)_

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -40,24 +40,20 @@ export const entityResource: Resource<Entity> = {
 };
 ```
 
-## Resource schemas mirror the platform — never tighten them
+## Client-side validation must match server-side validation
 
-A resource schema validates files the **platform** wrote, so the platform's own
-validator is the contract. Anything stricter here rejects an app that production
-accepts and evaluates, and because `readProjectConfig()` reads every resource up
-front, one such file fails *every* command — `deploy`, `entities push/pull`,
-`functions deploy` — whether or not it touches that resource.
+A resource schema validates files the server wrote and will read back, so the
+server's validation is the contract — never be stricter than it. Anything stricter
+rejects an app the server accepts, and because `readProjectConfig()` reads every
+resource up front, one such file fails *every* command (`deploy`,
+`entities push/pull`, `functions deploy`) whether or not it touches that resource.
+A hand-derived entity schema once rejected a quarter of real publishes this way.
 
-For entities the authorities are `backend/app/user_apps/entities/rls_validation.py`
-(`SUPPORTED_FIELD_OPERATORS`, `OPEN_RULE_VALUES`, `_user_condition_valid`),
-`backend/app/json_schema_utils.py` (`validate_json_schema`), and
-`backend/app/user_apps/common/entity_name_validation.py` (`ENTITY_NAME_PATTERN`)
-in the apper repo. Check against them before adding or narrowing a rule; a
-re-derived "obvious" rule is how `EntitySchema` came to reject a quarter of real
-publishes. Rejecting a shape the engine cannot evaluate (an unsupported operator,
-an operator inside `user_condition`) is correct — that's a real defect, not
-strictness. `base44 dev` interprets the same files locally, so a schema change
-usually needs a matching change in `src/cli/dev/dev-server/db/`.
+So before adding or narrowing a rule, confirm the server rejects that shape too.
+Rejecting something the server cannot evaluate — an unsupported operator, or an
+operator where only exact equality is applied — is correct; that's a real defect,
+not strictness. `base44 dev` interprets these same files locally, so a schema
+change usually needs a matching change in `src/cli/dev/dev-server/db/`.
 
 ## Adding a New Resource
 
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -40,6 +40,25 @@ export const entityResource: Resource<Entity> = {
 };
 ```
 
+## Resource schemas mirror the platform — never tighten them
+
+A resource schema validates files the **platform** wrote, so the platform's own
+validator is the contract. Anything stricter here rejects an app that production
+accepts and evaluates, and because `readProjectConfig()` reads every resource up
+front, one such file fails *every* command — `deploy`, `entities push/pull`,
+`functions deploy` — whether or not it touches that resource.
+
+For entities the authorities are `backend/app/user_apps/entities/rls_validation.py`
+(`SUPPORTED_FIELD_OPERATORS`, `OPEN_RULE_VALUES`, `_user_condition_valid`),
+`backend/app/json_schema_utils.py` (`validate_json_schema`), and
+`backend/app/user_apps/common/entity_name_validation.py` (`ENTITY_NAME_PATTERN`)
+in the apper repo. Check against them before adding or narrowing a rule; a
+re-derived "obvious" rule is how `EntitySchema` came to reject a quarter of real
+publishes. Rejecting a shape the engine cannot evaluate (an unsupported operator,
+an operator inside `user_condition`) is correct — that's a real defect, not
+strictness. `base44 dev` interprets the same files locally, so a schema change
+usually needs a matching change in `src/cli/dev/dev-server/db/`.
+
 ## Adding a New Resource
 
 1. Create folder: `packages/cli/src/core/resources/<name>/`
--- a/packages/cli/src/cli/dev/dev-server/db/rls.ts
+++ b/packages/cli/src/cli/dev/dev-server/db/rls.ts
@@ -154,9 +154,6 @@ export function checkRLS(
   user: Record<string, unknown> | undefined,
 ): boolean {
   if (user?.is_service === true) return true;
-  // OPEN_RULE_VALUES in the platform's rls_validation.py: absent, null, "", {}
-  // and true all impose no restriction, so they must not fall through to the
-  // condition evaluator (which denies anonymous users before evaluating).
   if (rule === undefined || rule === null || rule === "") return true;
   if (typeof rule === "object" && Object.keys(rule).length === 0) return true;
   if (typeof rule === "boolean") return rule;
--- a/packages/cli/src/cli/dev/dev-server/db/rls.ts
+++ b/packages/cli/src/cli/dev/dev-server/db/rls.ts
@@ -149,12 +149,16 @@ function evaluateCondition(
  * - condition object → evaluate against record and user
  */
 export function checkRLS(
-  rule: boolean | Record<string, unknown> | undefined,
+  rule: boolean | Record<string, unknown> | null | "" | undefined,
   record: Record<string, unknown>,
   user: Record<string, unknown> | undefined,
 ): boolean {
   if (user?.is_service === true) return true;
-  if (rule === undefined) return true;
+  // OPEN_RULE_VALUES in the platform's rls_validation.py: absent, null, "", {}
+  // and true all impose no restriction, so they must not fall through to the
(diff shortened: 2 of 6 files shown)


packages/cli/src/cli/dev/dev-server/db/rls.ts
```diff
@@ -149,12 +149,16 @@ function evaluateCondition(
...
  */
 export function checkRLS(
-  rule: boolean | Record<string, unknown> | undefined,
+  rule: boolean | Record<string, unknown> | null | "" | undefined,
   record: Record<string, unknown>,
   user: Record<string, unknown> | undefined,
 ): boolean {
   if (user?.is_service === true) return true;
-  if (rule === undefined) return true;
+  // OPEN_RULE_VALUES in the platform's rls_validation.py: absent, null, "", {}
```
netanelgilad: Drop the comment

packages/cli/src/cli/dev/dev-server/db/validator.ts
```diff
@@ -35,6 +35,9 @@ const fieldTypes = [
   "boolean",
   "array",
   "object",
+  // Only meaningful inside a union (`["string", "null"]`), which is how a
```
netanelgilad: Drop the comment

packages/cli/src/core/resources/entity/schema.ts
```diff
@@ -1,34 +1,23 @@
 import { z } from "zod";
 import { ResourceSourceSchema } from "@/core/resources/types.js";
 
+// This file mirrors the platform's contract; it must never be stricter. The
```
netanelgilad: Drop the comment

packages/cli/src/core/resources/entity/schema.ts
```diff
@@ -1,34 +1,23 @@
...
   z.string(),
   z.number(),
   z.boolean(),
   z.null(),
-  z.looseObject({
-    $in: z.unknown().optional(),
-    $nin: z.unknown().optional(),
-    $ne: z.unknown().optional(),
-    $all: z.unknown().optional(),
-  }),
```
netanelgilad, line 14: This too loose nope? Why this?

packages/cli/src/core/resources/entity/schema.ts
```diff
@@ -1,34 +1,23 @@
...
-    id: z.string().optional(),
-  })
-  .refine(
-    (val) =>
-      Object.keys(val).every(
-        (key) => userConditionAllowedKeys.has(key) || key.startsWith("data."),
-      ),
-    "Keys must be role, email, id, or match data.* pattern",
-  );
+// The engine compares user attributes by exact equality; it constrains neither
```
netanelgilad: Drop the comment. And is this too loose?

packages/cli/src/core/resources/entity/schema.ts
netanelgilad: Yes, too loose — and it already was, before this PR. Fixed in `b48bfa3`.

`z.looseObject({ $in: …, $nin: …, $ne: …, $all: … })` accepts **unknown keys** by definition, so the old declared-key list validated nothing; `created_by: {$contains: 1}` parsed fine on `main`. My `z.looseObject({})` was behaviourally identical, just more honestly written.

It matters more than it looks, because the node-level refine short-circuits on allowlisted *keys*:

```ts
rlsConditionAllowedKeys.has(key) || isValidFieldCondition(value)
```

`created_by` is in that set, so the refine returns true without ever looking at its value — leaving `FieldConditionSchema` as the only check. So the operator allowlist was enforced for `data.x: {…}` but not for the built-in fields.

It now refines with the same `isValidFieldCondition`, so both paths agree: `created_by: {$gte: "…"}` passes, `created_by: {$contains: "x"}` fails. Test: "enforces the operator allowlist on built-in fields too".

---

packages/cli/src/core/resources/entity/schema.ts
netanelgilad: Comment dropped, and yes — too loose. Fixed in `b48bfa3`.

I'd gone to bare `looseObject({})` to avoid adding any new rejection, but that overshot: it accepts things the engine provably cannot evaluate. It now mirrors `_user_condition_valid` — a **non-empty** object of `attribute: scalar` pairs, **no operators**:

```ts
Object.keys(val).length > 0 &&
  Object.entries(val).every(([key, value]) => !key.startsWith("$") && isScalar(value))
```

`user_condition` does exact equality, so `{role: {$in: [...]}}` or `{role: ["a"]}` matches nothing at runtime and **blocks all access** — fail-closed, but still a bug worth catching at parse time rather than shipping.

What stays dropped is the old `role`/`email`/`id`/`data.*` **key** allowlist, which is the part that caused real failures — the engine doesn't restrict which user attributes you compare, so `{tenant_id: "acme", is_staff: true}` is fine and used to be fatal.

One rule from `_user_condition_valid` I deliberately left out: it also rejects `{{user.x}}` templates inside `user_condition` values (they aren't resolved there, so they never match). That one would newly break files that parse today and look natural, so I'd rather it be its own change than a side effect of this one — say the word and I'll add it.

---

packages/cli/src/core/resources/entity/schema.ts
netanelgilad: Dropped in `b48bfa3`. The one thing it carried that I didn't want to lose — "mirror the platform, never tighten", plus which server files are the authority — moved to `docs/resources.md`, where someone adding a resource schema will actually run into it.

---

packages/cli/src/cli/dev/dev-server/db/rls.ts
netanelgilad: Dropped in `b48bfa3` — the `checkRLS` predicate reads as its own explanation.

---

packages/cli/src/cli/dev/dev-server/db/validator.ts
netanelgilad: Dropped in `b48bfa3` (both this one and the union-handling comment a few lines down).

---

docs/resources.md
```diff
@@ -40,6 +40,25 @@ export const entityResource: Resource<Entity> = {
...
 
+## Resource schemas mirror the platform — never tighten them
+
+A resource schema validates files the **platform** wrote, so the platform's own
+validator is the contract. Anything stricter here rejects an app that production
+accepts and evaluates, and because `readProjectConfig()` reads every resource up
+front, one such file fails *every* command — `deploy`, `entities push/pull`,
+`functions deploy` — whether or not it touches that resource.
+
+For entities the authorities are `backend/app/user_apps/entities/rls_validation.py`
```
netanelgilad: Don't mention backend project paths or anything just say that client side validation has to match server side validation

netanelgilad: Done in `7b5c886`. The section is now just the rule and why it bites here:

> **Client-side validation must match server-side validation**
> A resource schema validates files the server wrote and will read back, so the server's validation is the contract — never be stricter than it. […] because `readProjectConfig()` reads every resource up front, one such file fails *every* command whether or not it touches that resource.

No backend paths or symbol names — you're right that they'd rot against another repo's layout, and the principle is what a reader needs anyway. I kept the one in-repo pointer (`src/cli/dev/dev-server/db/`), since a schema change usually needs a matching change there.

I also scrubbed the same references out of the two new specs, which had picked up a few server symbol names in test titles and comments.

Left the PR description as-is since it's review context rather than committed content — happy to trim it too if you'd rather it not sit in the history.

---

guyofeck (approved):
LGTM

PR #595, opened 2026-08-11
feat(logs): true realtime `--follow` via apper SSE stream (CF tail) — DRAFT

davidsu:
> [!NOTE]
> ## Description
>
> Turns `base44 logs --follow` from a 2-second poll over an eventually-consistent index (~20-30s lag) into a consumer of the realtime SSE log stream (`GET /api/apps/{app_id}/functions-mgmt/logs/stream`), measured at ~0.65s end-to-end vs ~17s on the poll path. The stream is read by hand over `fetch` (EventSource cannot set auth headers), validated with Zod, and driven by a typed `event: end` payload that tells the CLI whether a disconnect is retriable. The command decides once, at connect time, whether streaming is possible: if the endpoint is unavailable it says so on stderr and runs the old poll loop for the whole session; if a stream was established and later cannot be re-established, it exits with an actionable error rather than silently degrading mid-tail.
>
> ## Related Issue
>
> None (companion backend PR: base44-dev/apper#19705)
>
> ## Type of Change
>
> - [ ] Documentation update
> - [x] Other (please describe): one deliberate v1 downscope — `--since` combined with `--follow` is now rejected instead of silently backfilling
>
> ## Changes Made
>
> - **New `core/resources/function/stream-api.ts`**: `openLogStream(filters)` opens the SSE endpoint with the same auth as the bounded logs route (`Authorization: Bearer <jwt>`, or the `api_key` header for workspace keys, with proactive token refresh) and returns a `LogStreamAttempt` — either an async generator of Zod-validated `StreamEvent`s (`log` | `end` | `ping`), a `refused`, or a `transient`. `parseStreamEvent`, `readStreamEvents`, and `isWorthReconnecting` are exported for unit testing.
> - **SSE reader**: tracks `event:` names so the typed `end` payload parses against its own schema instead of silently failing the log schema; `:` comment lines surface as `ping` events; unknown event names and malformed payloads are skipped so one bad line cannot kill a live tail. Stream `level: "warn"` is normalized to `warning`, matching the bounded route, so `--level warning` filters both paths identically.
> - **Bounded silent-failure modes**: a 10s connect-phase timeout on the stream fetch (headers only) and a 60s line-silence watchdog on the body — keepalive pings reset it, so quiet-but-healthy apps never trip it. Worst-case mute is bounded at ~70s.
> - **Keepalive pings count as liveness**: a connection that delivered only `:` ping comments is treated as proven alive (`provedAlive`), so an idle app's stream keeps reconnecting instead of burning the give-up budget.
> - **Transient connect retries**: an unreachable or 5xx endpoint is retried with backoff (1s/2s/4s/8s) before conceding; a deliberate refusal (4xx) is taken at face value immediately. `isWorthReconnecting(status)` draws that line at 500.
> - **Reason-driven reconnect policy in `logs.ts`**: `retriable: true` reconnects after 1s; `retriable: false` stops; a bare EOF/error with no `end` event reconnects and gives up only after 2 consecutive drops that produced zero events (the counter resets whenever a stream proves alive, so long sessions never exhaust a budget).
> - **One decision, then commit to it**: `followLogs` connects first. On failure it warns on stderr (`refused` → "not available for this app", `transient` → "could not reach") and hands off to `pollLogs` for the rest of the session. Once a stream is live, exhausting the reconnect budget throws an `ApiError` with hints (`base44 logs --follow`, `base44 logs`) instead of quietly switching transports behind the user's back.
> - **Poll path preserved**: the old loop is extracted as `pollLogs` with a seedable cursor and unchanged behavior (2s interval, `selectNewEntries` dedup).
> - **`--since` + `--follow` now errors** like `--until`/`--order` — the seam between the lagging bounded index and a tail-from-connect stream left a silent data hole, and closing it properly needs poll-until-caught-up splicing. `--since` alone is unchanged.
> - **`--json --follow` still emits NDJSON** with the exact same `LogEntry` shape (`time`, `level`, `message`, `source`) as the poll path.
> - **`--help` no longer promises a `--limit` default that never existed**: the text now reads `Results per page (1-1000; the server returns at most 500)`. The CLI has never sent a default limit, so the old `default: 50` was a documentation-only bug already on `main`.
>
> ## Testing
>
>
> Added 173 lines of tests in `tests/cli/logs.spec.ts`: the SSE parser (valid log event, `warn`→`warning`, `function: null` kept rather than dropped, unknown event names and malformed payloads ignored, typed `end` parsing), keepalive-comment handling in `readStreamEvents`, the `printStreamUntilEnd` liveness signal, the drop-budget arithmetic (including a silent drop following a proven one), and the reconnect-vs-refuse status classification — plus CLI specs for the `--since`/`--follow` rejection and the corrected `--limit` help text. The live stream itself is not child-process integration-testable (`--follow` never exits) and was verified with a two-terminal demo against a local backend. The suite was **not** re-run while generating this description — `bun run build` and `vitest` were both denied in this environment. Previously on this branch `logs.spec.ts` was green and the only failures were in `exec.spec.ts`, caused by a local npm-registry TLS interception (Deno cannot resolve `@base44/sdk`) and unrelated to this diff.
>
> ## Checklist
>
>
> ## Additional Notes
>
> - Marked DRAFT: it depends on the companion backend endpoint landing first. Workspace API keys currently 403 on logs routes, and the stream inherits that limitation.
> - `--level` stays client-side filtered, as on the poll path; only `function` and `env` are applied server-side.
> - No event ids or resume in v1, no backpressure handling, no config surface — deliberately minimal.
> - Diff is 4 files, +538/−4: `logs.ts`, the new `stream-api.ts`, its barrel export, and `tests/cli/logs.spec.ts`.
>
> ---
> <sub>🤖 Generated by Claude | 2026-08-31 09:20 UTC | e3f5cae</sub>

### DIFF
--- a/packages/cli/src/cli/commands/project/logs.ts
+++ b/packages/cli/src/cli/commands/project/logs.ts
@@ -9,6 +9,7 @@ import type {
   FunctionLogsResponse,
   LogEnv,
   LogLevel,
+  LogStreamAttempt,
   LogStreamFilters,
   StreamEndEvent,
   StreamEvent,
@@ -170,6 +171,19 @@ async function printStreamUntilEnd(
 
 const STREAM_RECONNECT_DELAY_MS = 1_000;
 const MAX_DROPS_SINCE_LAST_EVENT = 2;
+const CONNECT_RETRY_DELAYS_MS = [1_000, 2_000, 4_000, 8_000];
+
+async function connectWhileTransientlyUnavailable(
+  filters: LogStreamFilters,
+): Promise<LogStreamAttempt> {
+  let attempt = await openLogStream(filters);
+  for (const retryDelay of CONNECT_RETRY_DELAYS_MS) {
+    if (attempt.kind !== "transient") return attempt;
+    await delay(retryDelay);
+    attempt = await openLogStream(filters);
+  }
+  return attempt;
+}
 
 interface StreamAttemptResult {
   everConnected: boolean;
@@ -188,11 +202,11 @@ async function streamUntilExhausted(
   let lastTime = "";
   let dropsSinceLastEvent = 0;
   while (true) {
-    const stream = await openLogStream(filters);
-    if (!stream) break;
+    const attempt = await connectWhileTransientlyUnavailable(filters);
+    if (attempt.kind !== "stream") break;
     everConnected = true;
     const ending = await printStreamUntilEnd(
-      stream,
+      attempt.events,
       options.level,
       jsonMode,
       lastTime,
--- a/packages/cli/src/cli/commands/project/logs.ts
+++ b/packages/cli/src/cli/commands/project/logs.ts
@@ -194,46 +194,53 @@ async function connectWhileTransientlyUnavailable(
   return attempt;
 }
 
-interface StreamAttemptResult {
-  everConnected: boolean;
-  lastTime: string;
-}
-
 async function streamUntilExhausted(
+  firstStream: AsyncGenerator<StreamEvent>,
+  filters: LogStreamFilters,
   options: LogsOptions,
   jsonMode: boolean,
-): Promise<StreamAttemptResult> {
-  const filters: LogStreamFilters = {
-    functions: parseFunctionNames(options.function),
-    env: options.env,
-  };
-  let everConnected = false;
+): Promise<void> {
+  let events = firstStream;
   let lastTime = "";
   let dropsSinceLastEvent = 0;
   while (true) {
-    const attempt = await connectWhileTransientlyUnavailable(filters);
-    if (attempt.kind !== "stream") break;
-    everConnected = true;
     const ending = await printStreamUntilEnd(
-      attempt.events,
+      events,
       options.level,
       jsonMode,
       lastTime,
     );
     lastTime = ending.lastTime;
     if (ending.end) {
-      if (!ending.end.retriable) break;
+      if (!ending.end.retriable) return;
       dropsSinceLastEvent = 0;
     } else {
       dropsSinceLastEvent = countDropTowardGivingUp(
         dropsSinceLastEvent,
         ending.provedAlive,
       );
-      if (shouldGiveUpStreaming(dropsSinceLastEvent)) break;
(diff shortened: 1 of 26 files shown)


guyofeck (commented):
Read the diff, ran `typecheck`/`lint`/`knip`/`logs.spec.ts` locally (49/49 green). Shape is right — three-way `LogStreamAttempt`, typed `end` driving retry, polling kept as a live fallback. Three things before this leaves draft:

**1. Body never cancelled** — `stream-api.ts:137`: the `finally` calls `releaseLock()` but not `reader.cancel()`. Verified with a spied `ReadableStream`: breaking out on `event: end` never cancels, so each reconnect (and the final drop into the forever-running `pollLogs`) abandons an open connection. `try { await reader.cancel(); } catch {}` — cancel also releases the lock.

**2. Auth errors classified as `transient`** — `stream-api.ts:183-193`: `buildStreamAuthHeaders()` is awaited inside the `fetch(...)` args, so a logged-out user's `readAuth()` throw is caught as `transient` and retried 1+2+4+8s before the poll path surfaces the real error. Repro: `base44 logs -f --function foo` while logged out. Build the headers and URL before the `try`.

**3. `--since` + `--follow` regression** — on `main` the first poll passed `since` through, so backfill-then-tail worked; now it hard-errors while "Type of Change" says non-breaking. One `since`-bounded fetch before opening the stream would keep it (`selectNewEntries` already dedupes) — otherwise check the breaking box.

Nice drive-by on the `--limit` help text.

---

packages/cli/src/cli/commands/project/logs.ts
```diff
@@ -122,14 +129,156 @@ function writeFollowLine(entry: LogEntry, jsonMode: boolean): void {
   process.stdout.write(`${line}\n`);
 }
 
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
netanelgilad, line 132: this is now native in node if you import setTimeout from node:timers/promises

netanelgilad (approved):

PR #597, opened 2026-08-11
ci: Bun cooldown for installs, no gateway in publish, OIDC

guyofeck:
Implements the interim policy @dimar set out in [this thread](https://base44workspace.slack.com/archives/C095K108GKU/p1786444151352039?thread_ts=1786436550.460409), so Actions can be re-enabled for `base44/cli`.

| Requirement | How it's met |
|---|---|
| 1. Lock-files, build never overrides one | `bun install --frozen-lockfile`, already in all 9 install steps — unchanged |
| 2. Package manager with a minimal-age directive | `minimumReleaseAge = 1209600` (14 days) in new `bunfig.toml` |
| 3. Embargo in non-publish tasks only | Gateway removed from both publish workflows; `/etc/hosts` unpin hack deleted |

**No package manager migration needed — Bun supports this natively:** `bun install --minimum-release-age=`. @dimar, your list said "pnpm/yarn/(latest)npm"; Bun belongs on it.

Dropping the gateway from publish is safe because those jobs resolve nothing — `--frozen-lockfile` installs `bun.lock` verbatim. `check_wix_proxy_steps.py` keeps the proxy mandatory everywhere else: the two publish workflows are exempted by exact filename in a `PUBLISH_WORKFLOWS` list, and every exemption is printed on success, so the list cannot quietly grow.

Also: publishes authenticate via trusted publishing (OIDC), so `NPM_TOKEN` is gone; and `bunx json-bump` becomes `npm version` / `npm pkg set`, since `bunx` fetched an undeclared package at run time and doesn't enforce the cooldown ([#30748](https://github.com/oven-sh/bun/issues/30748)).

### Why the gateway can't just carry the publish

`npm publish` sends `PUT /<package>`, which matches neither the embargo's `^~ /-/` passthrough block nor `~ \.tgz$`, so it lands in `location /` (proxy_metadata, a read path with caching). The gateway also sets no `client_max_body_size`, so nginx's 1 MB default rejects a packument carrying the base64 tarball. Worth fixing on the embargo side — once it is, these two workflows should go back to being gatewayed like everything else.

Bun's own gaps, for the record: `bunx` ignores `--minimum-release-age` ([#30748](https://github.com/oven-sh/bun/issues/30748)), `bun update --latest` skips the cooldown for transitive deps ([#25305](https://github.com/oven-sh/bun/issues/25305)), and there's no bypass when an update fixes a known vulnerability ([#26065](https://github.com/oven-sh/bun/issues/26065)) — `minimumReleaseAgeExcludes` is the escape hatch there.

## Status

- [x] `check_wix_proxy_steps.py`: 12/12 non-publish jobs across 11 workflows, 2 publish workflows exempt; 11/11 unit tests
- [ ] **Nothing here has run** — Actions is disabled org-wide
- [ ] Trusted publishers on npmjs.com — needs npm org admin, one entry per workflow filename

## Questions for @dimar

1. Does Bun's `minimumReleaseAge` satisfy requirement 2? Set to 14 days here — say if you want a different window.
2. Nothing verifies a committed lockfile change respected the cooldown (CI never resolves). Want a check for that, and for `--frozen-lockfile` itself?

---


.github/scripts/check_wix_proxy_steps.py
```diff
@@ -118,6 +143,20 @@ def job_problem(job: dict, workflows: frozenset[str]) -> str | None:
...
+    if "uses" in job:
+        return None
+    steps = job.get("steps") or []
+    if any(_uses(step) == PROXY_ACTION for step in steps):
+        return (
+            "runs the Wix gateway proxy, but publish workflows must not: the gateway "
+            "cannot carry `npm publish`, so these workflows rely on the committed "
+            "lockfile plus min-release-age in .npmrc instead"
+        )
+    return None
```
Paveltarno: not sure this is really needed, maybe its enough to simply not check the publishing actions set explicitly in the list?

bunfig.toml
```diff
@@ -0,0 +1,24 @@
...
+# bun.lock verbatim and never resolves, so it is unaffected by design.
+#
+# Known gaps, accepted knowingly rather than discovered later:
+#   - `bunx` accepts --minimum-release-age but does not enforce it
+#     (oven-sh/bun#30748). Do not fetch tooling with bunx in a release path;
+#     the publish workflows use `npm version` / `npm pkg set` instead.
+#   - `bun update --latest` skips the cooldown for transitive deps
+#     (oven-sh/bun#25305).
+#   - No bypass when an update fixes a known vulnerability
+#     (oven-sh/bun#26065) — use minimumReleaseAgeExcludes for that case.
```
Paveltarno: let's delete this slop

.github/scripts/check_wix_proxy_steps.py
```diff
@@ -19,6 +22,28 @@
...
+# minimal-age directive (`minimumReleaseAge` in bunfig.toml) instead. Those jobs
+# resolve nothing, so dropping the gateway does not widen what they can pull.
+# The gateway cannot carry a publish today — `npm publish` sends
+# `PUT /<package>`, which misses its `^~ /-/` passthrough block, and it sets no
+# client_max_body_size so nginx's 1 MB default rejects a packument carrying the
+# base64 tarball.
+#
+# Keyed on exact filename, and every exemption is printed on success, so this
+# cannot quietly grow. Revisit once the embargo publish bug is fixed: these
+# workflows should go back to being gatewayed like everything else.
```
Paveltarno: lets remove slop, enough to write that adding files here should not be taken easily and can pose a security concern

Paveltarno (changes requested):

Paveltarno (dismissed):
לגטמ

Paveltarno (approved):

PR #602, opened 2026-08-17
site deploy: finalize through the session id the create returned

netanelgilad:
> [!NOTE]
> ## Description
>
> A deployment id is derived from the commit, so two deploys of the same commit address the same deployment — the loser of a create race finalizes against whatever the winner's uploads left behind. The platform now returns a per-attempt `session_id` from `POST deployments`; this PR threads it back as `?session_id=` on finalize so each run resolves its own uploads. Pairs with base44-dev/apper#20445 on the platform side.
>
> ## Related Issue
>
> Pairs with base44-dev/apper#20445 (platform side). No GitHub issue in this repo.
>
> ## Type of Change
>
> - [x] Documentation update
>
> ## Changes Made
>
> - `CreateDeploymentResponseSchema` gains a **required** `session_id`, transformed to `sessionId` alongside `deploymentId` (`packages/cli/src/core/site/schema.ts`).
> - `finalizeStaticDeployment` and `postFinalize` now take a `sessionId` and send it as `searchParams: { session_id }` on `POST deployments/{id}/finalize` (`packages/cli/src/core/site/api.ts`).
> - `deployStaticSite` threads `created.sessionId` from create through to finalize (`packages/cli/src/core/site/static-site.ts`).
> - `TestAPIServer` captures finalize query strings in a new `finalizeQueries` array (it previously captured only multipart fields), and its `DeploymentCreateResponse` shape carries `session_id`.
> - `static_site_deployments.spec.ts` mocks a session id on create and asserts finalize sends it: `expect(t.api.finalizeQueries[0]).toEqual({ session_id: SESSION_ID })`.
> - `docs/deployments.md` updated to document `session_id` in the create response, the `?session_id=` finalize param and why it exists, and the new `finalizeQueries` testkit capture.
>
> ### Why \`session_id\` is required, not optional
>
> An earlier revision made it optional so a newer CLI would degrade against an older platform. There is nothing to tolerate: the static lane is gated by \`BASE44_STATIC_DEPLOYMENTS\`, set in exactly one place on the platform (the sandbox publish path) behind the currently-closed \`sandbox-cli-static-publish\` flag, and nothing else calls the deployments API. Public \`base44 site deploy\` users take the legacy tar.gz path and never reach this code.
>
> ## Testing
>
>
> ## Checklist
>
> - [x] I have updated \`docs/\` (AGENTS.md) if I made architectural changes
>
> ## Additional Notes
>
> Recorded run from when the change landed on the branch: \`bun run test static_site_deployments\` — 14 passed. It was not re-run for this description pass. The session assertion lives in the existing happy-path deploy test rather than a standalone case, which would have been setup-identical. Making \`session_id\` required also forced every mocked create in the spec to carry one — the schema doing its job.
>
> Full-suite caveats from that same run, both verified **pre-existing on clean \`main\`** and not introduced here: \`bun run typecheck\` reports 6 errors in \`src/cli/dev/dev-server/function-bundler.ts\` (identical count with the change stashed), and \`bun run test\` fails 25 tests in \`dev.spec.ts\` / \`exec.spec.ts\` because \`deno\` is not installed in that environment.
>
> ---
> <sub>🤖 Generated by Claude | 2026-08-30 11:57 UTC | c0623af</sub>


packages/cli/src/core/site/api.ts
```diff
@@ -84,26 +84,34 @@ export async function createDeployment(
...
 
   let response: KyResponse;
   try {
     response = await appClient.post(
       `deployments/${encodeURIComponent(deploymentId)}/finalize`,
-      { body: formData, timeout: 180_000 },
+      {
+        body: formData,
+        timeout: 180_000,
+        // Resolves this attempt's uploads rather than a concurrent deploy's:
```
netanelgilad: Remove the comment

packages/cli/tests/cli/testkit/TestAPIServer.ts
```diff
@@ -237,6 +237,9 @@ interface CreateAppResponse {
 
 interface DeploymentCreateResponse {
   deployment_id: string;
+  /** This attempt's upload session. Optional here so a spec can mock a
```
netanelgilad: No need to test a scenario like that, it won't happen the platform will always return it

packages/cli/src/core/site/api.ts
netanelgilad: Removed. `searchParams: { session_id: sessionId }` reads for itself, and the why belongs in the commit message, not on the line.

---

packages/cli/tests/cli/testkit/TestAPIServer.ts
netanelgilad: Agreed — dropped. `session_id` is required on the testkit type again, and the `fails clearly when the platform opens no upload session` spec is gone. The only reason it was optional was to mock that case, so both went together.

One thing I did beyond the comment, push back if you disagree: with the negative case gone, `finalizes through the session the create handed back` was setup-identical to `sends no PUTs and still finalizes when every asset is already stored` and differed only in its one assertion. Rather than leave a near-duplicate, I folded `expect(t.api.finalizeQueries[0]).toEqual({ session_id: SESSION_ID })` into the happy-path deploy test and deleted the standalone one. Same coverage, one fewer test — 14 passing.

---

DolevEpshtein (approved):

Hustleofficial:
Rcb

PR #607, opened 2026-08-27
fix(logs): avoid loading local project resources

davidsu:
> [!NOTE]
> ## Description
>
> `base44 logs` used to read the local project config to figure out which functions to fetch logs for. That meant an unrelated local config error (e.g. an invalid entity file) could break a purely remote, read-only command, and the "function not found" hint listed local functions rather than what is actually deployed. This PR makes the command source function names exclusively from the deployed-functions API, and only when it actually needs them.
>
> It also hardens the 404 path: the available-functions hint is now looked up lazily, and if that lookup fails the original logs error is rethrown instead of being masked by a secondary failure. When the hint does succeed, the 404 is attached as the `cause` of the resulting `InvalidInputError`.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - Dropped `readProjectConfig` / `getProjectFunctionNames` from `logs.ts` — the command no longer loads local project resources, so a broken local config cannot fail a remote log fetch.
> - Function names are now resolved only from `listDeployedFunctions()`, and only when `--function` is omitted; with `--function` specified no list call is made at all.
> - Added `getFunctionNamesForHint()`: on a 404 it lazily fetches deployed function names for the hint, and rethrows the original logs error if that fetch fails, so a secondary error never masks the real one.
> - The `InvalidInputError` for a missing function now carries the 404 `ApiError` as its `cause`, and the hint text reads "Available functions in this app" instead of "in this project".
> - Unified the empty-state message to "No functions found in this app." (no more local/remote split).
> - Updated the `--function` help text to say logs are fetched for "all deployed functions" when the flag is omitted.
> - Tests: added coverage for skipping local resources with `--function`, listing deployed names in the 404 hint, and preserving the original error when the hint lookup fails; reworked existing tests to mock the functions list and removed now-unnecessary mocks.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> Behavior change worth noting: inside a local project, `base44 logs` with no `--function` now lists logs for functions deployed to the app rather than functions declared locally. Functions that exist locally but have never been deployed will no longer appear — which matches what the logs API can actually return.
>
> The test suite was not executed in this environment (running it needs `bun run build` first), so the "tested locally"/"all tests pass" boxes are left unchecked pending CI.
>
> ---
> <sub>🤖 Generated by Claude | 2026-08-27 10:26 UTC | 4743740</sub>

### DIFF
--- a/packages/cli/src/cli/commands/project/logs.ts
+++ b/packages/cli/src/cli/commands/project/logs.ts
@@ -3,7 +3,6 @@ import { Option } from "commander";
 import type { CLIContext, RunCommandResult } from "@/cli/types.js";
 import { Base44Command, normalizeDatetime } from "@/cli/utils/index.js";
 import { ApiError, InvalidInputError } from "@/core/errors.js";
-import { readProjectConfig } from "@/core/index.js";
 import type {
   FunctionLogFilters,
   FunctionLogsResponse,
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli/commands/project/logs.ts
```diff
@@ -188,27 +187,29 @@ async function fetchLogsForFunctions(
...
-          },
-        );
+      if (error instanceof ApiError && error.statusCode === 404) {
+        const namesForHint =
+          availableFunctionNames.length > 0
+            ? availableFunctionNames
+            : await getRemoteFunctionNames();
+        if (namesForHint.length > 0) {
+          const available = namesForHint.join(", ");
+          throw new InvalidInputError(
```
talge-a11y, line 197: You are loosing the original error here.

talge-a11y (dismissed):
see my note

packages/cli/src/cli/commands/project/logs.ts
davidsu: You are right. The lazy function-list lookup could replace the original logs 404 if that secondary request failed. Updated it to rethrow the original logs error when hint loading fails, with a regression test covering both error messages. Fixed in 5fdb7ee.

talge-a11y (approved):

PR #609, opened 2026-08-31
fix(site): drop the per-file asset size limit from the static manifest

netanelgilad:
> [!NOTE]
> ## Description
>
> Removes the 25 MiB per-file size limit that `buildAssetManifest()` enforced on static assets. The limit existed because assets were read fully into memory via `readFile()` before hashing; hashing now streams the file in chunks through `createReadStream()`, so an asset is never held whole in memory at manifest time and there is no reason to reject large files. Deployments containing large assets (videos, WASM bundles, model files) no longer fail at manifest build.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [x] Documentation update
>
> ## Changes Made
>
> - Dropped `MAX_ASSET_SIZE_BYTES` and the `InvalidInputError` thrown for oversized assets in `packages/cli/src/core/site/manifest.ts`.
> - Added an internal `hashAssetFile()` helper that computes the same app-id-salted sha256 digest as `hashAsset()` but consumes the file as a read stream, so memory use is bounded by the chunk size rather than the file size.
> - `buildAssetManifest()` still `stat()`s each file for the manifest `size` field, but no longer reads its contents into a `Buffer`.
> - Updated `docs/deployments.md`: replaced the "files over 25 MiB fail with a per-file error" note with a statement that there is no per-file limit and that assets are hashed by streaming; the 100,000-file cap is unchanged.
> - Tests: replaced the oversized-file rejection test with one asserting a >25 MiB file lands in the manifest with the correct `size`, and added a test that a multi-chunk (200 KiB) file hashes to the same digest as `hashAsset()` over the whole buffer, guarding against chunk-boundary regressions.
> - Kept `hashAssetFile()` unexported (Knip flags unused exports) and formatted the new assertions per Biome.
>
> ## Testing
>
>
> ## Checklist
>
>
> ## Additional Notes
>
> The hash format is unchanged — `sha256(utf8(app_id) || raw file bytes)` truncated to 32 hex chars — so previously deployed manifests and server-side dedup by hash remain valid. Upload of large assets is still subject to whatever limits the presigned-URL storage backend imposes; this change only removes the client-side manifest-time rejection.
>
> ---
> <sub>🤖 Generated by Claude | 2026-08-31 09:43 UTC | cbdaa40</sub>


netanelgilad:
@claude fix the failing PR checks here

davidsu (approved):

PR #616, opened 2026-09-10
Support sandbox branch targeting by name

theunreal:
Sandbox commands target a Base44 branch by name with `--branch <name>`. For example:

```sh
base44 --app-id <app-id> --branch feature/checkout sandbox read src/App.jsx
```

The CLI resolves the exact name within the selected app and sends the ID to the sandbox bridge internally. Missing or ambiguous names fail before any sandbox operation. `--branch main` or no flag targets main. No checkout state is saved, and there is no public ID flag.

All seven sandbox commands honor the global flag; unsupported commands such as functions pull/list reject it explicitly. Broader resource branch support remains separate work.

Validation: production bundle build, TypeScript typecheck, lint, Knip, and 37 mocked-HTTP CLI tests passed. No live E2E run.

Pairs with base44-dev/apper#23840, whose Builder-generated CLI instructions now include the selected branch name. MCP sandbox tools are an independent connection and do not invoke the CLI.

Branch discovery: `base44 --app-id <app-id> branches list` returns main and active branch names, with structured JSON output for agents; IDs remain internal.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -4,6 +4,19 @@
 
 Commands live in `src/cli/commands/<domain>/`. They use a **factory pattern** — each file exports a function that returns a `Base44Command`.
 
+## Branch targeting
+
+`--branch-id <id>` is global, but currently supported only by sandbox commands.
+For example: `base44 --branch-id <id> sandbox read <path> --app-id <app-id>`.
+Other commands (including `functions pull`, `functions list`, and `entities push`)
+reject it before authentication or command execution, rather than silently targeting main.
+There is no `entities pull` command; branch source files can be read through `sandbox read`.
+
+To add support, set `supportsBranch: true` on the command and consume
+`ctx.branchId`. First verify that every backend operation honors that scope;
+accepting the query parameter alone is not proof of branch isolation.
+Omitting the flag preserves existing main-app behavior.
+
 ## Command File Template
 
 ```typescript
--- a/packages/cli/src/cli/commands/sandbox/checkpoint.ts
+++ b/packages/cli/src/cli/commands/sandbox/checkpoint.ts
@@ -3,15 +3,14 @@ import type { CLIContext, RunCommandResult } from "@/cli/types.js";
 import { Base44Command } from "@/cli/utils/index.js";
 import { getAppContext } from "@/core/project/index.js";
 import { createCheckpoint } from "@/core/resources/sandbox/api.js";
-import type { SandboxBranchOptions } from "./shared.js";
 import { toJsonStdout } from "./shared.js";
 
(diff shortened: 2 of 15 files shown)


packages/cli/src/cli/commands/sandbox/checkpoint.ts
```diff
@@ -29,6 +33,7 @@ export function getSandboxCheckpointCommand(): Command {
       "--name <name>",
       "Optional message/title for the checkpoint (defaults to an auto-generated title)",
     )
+    .option("--branch-id <id>", "Operate on a specific app branch")
```
carmelc: I am wondering if only the the sandbox cli commands need it, like don't`functions pull/entity pull` (get the backend code and entity schemas) need the branch id too (I am not sure if it is covered in the feature), if it is relevant in more commands you can consider adding it to `packages/cli/src/cli/program.ts` (global params added to all cli commands)

carmelc (dismissed):

packages/cli/src/cli/commands/sandbox/checkpoint.ts
theunreal: Good point, I made --branch-id global. sandbox commands support it, while unsupported commands like functions pull explicitly reject it instead of silently targeting main.

docs/commands.md
```diff
@@ -4,6 +4,19 @@
 
 Commands live in `src/cli/commands/<domain>/`. They use a **factory pattern** — each file exports a function that returns a `Base44Command`.
 
+## Branch targeting
+
+`--branch-id <id>` is global, but currently supported only by sandbox commands.
```
netanelgilad: why is it called branch-id and not just branch? what is an example value that needs to be passed, is it really an id?

netanelgilad (dismissed):

docs/commands.md
```diff
@@ -4,6 +4,22 @@
 
 Commands live in `src/cli/commands/<domain>/`. They use a **factory pattern** — each file exports a function that returns a `Base44Command`.
 
+## Branch targeting
+
+`--branch <name>` and `--branch-id <id>` are global, but currently supported only by sandbox commands.
```
netanelgilad: we dont need both :) just branch.
users and agents have no way to know the id

docs/commands.md
theunreal: The CLI now accepts only --branch <name> and resolves the ID internally.

docs/commands.md
theunreal: The ID flag is removed, and branches list now lets agents discover names directly.

theunreal:
Moved unchanged to internal PR #622 so trusted publishing and review workflows can run.

PR #635, opened 2026-09-22
feat(cli): install and serve a site through the CLI

ronnyrin:
> [!NOTE]
> ## Description
>
> Adds the two commands a hosted build sandbox needs to bring a project up without a developer at the keyboard: `base44 install` (run the site's `installCommand` and nothing else) and `base44 site dev` (run the site's dev server with no local backend, the frontend reaching its backend same-origin). `base44 build` is also relaxed to `requireAuth: false`, since it resolves the app id locally and runs a local build. All three are local-only paths, so a machine that has never logged in can install, build and serve a checkout.
>
> `site dev` takes no arguments at all: the bind address and the dev server's spelling of its bind-address flag come from new `site.devHost` / `site.devPort` / `site.devHostFlag` config keys, which default to `0.0.0.0`, `5173` and `--host`. A `serveCommand` that cannot receive forwarded arguments is refused rather than silently served on a different address.
>
> ## Related Issue
>
> None
>
> ## Type of Change
>
> - [ ] Documentation update
>
> ## Changes Made
>
> - **New `base44 install`** (`src/cli/commands/project/install.ts`): runs the site block's `installCommand` via `runTask`, registered in `program.ts`. `requireAuth: false` and `requireAppContext: false` — no app to resolve, no API call. Fails with an actionable hint when the project has no `site` block.
> - **New `base44 site dev`** (`src/cli/commands/site/dev.ts`): reads the project config, falls back to `npm run dev` when the site block names no `serveCommand`, appends the configured bind address, and runs it through the existing `ServeCommandRunner`. `requireAuth: false`; exits with the child's code and tears down on SIGINT/SIGTERM.
> - **New `withServeAddress()` core helper** (`src/core/site/serve-command.ts`): appends `<hostFlag> <host> --port <port>` to `npm run <script>` (through `--`) and to `pnpm`/`yarn`/`bun run` scripts (directly), plus `npm --prefix <path> run <script>`. Script and prefix tokens are matched as narrow charsets rather than `\S+`, so shapes like `npm run dev;evil` or `npm run $(id)` are not treated as forwarding shapes. Returns `droppedAddress` alongside the command so "no address to append" cannot be confused with "the address went nowhere"; `site dev` turns a dropped address into a hard failure instead of a warning plus exit 0.
> - **Config schema** (`src/core/project/schema.ts`): `site.devHostFlag` (default `--host`; `--hostname` for Next, which exits on `--host`), `site.devHost` (default `0.0.0.0`) and `site.devPort` (default `5173`, validated 1–65535). `serveCommand` and `outputDirectory` still deliberately do not default — their absence is the signal `base44 dev` and `deploy` read.
> - **`base44 build` no longer requires auth** (`src/cli/commands/project/build.ts`): `requireAuth: false`, since the app id comes from `--app-id`, `BASE44_APP_ID` or `.app.jsonc` and the build is local.
> - **`appBaseUrl` is now optional on `ServeCommandRunnerOptions`**: when omitted, `VITE_BASE44_APP_BASE_URL` is not injected, so a same-origin frontend (the Base44 vite plugin proxies `/api`) is not pointed across origins. `base44 dev` still passes it.
> - **Extracted `stopRunnerOnProcessSignals()`** out of `commands/dev.ts` into `src/cli/dev/stop-runner-on-signals.ts` so both `dev` and `site dev` share it.
> - **CHANGELOG** updated with entries for all of the above.
>
> ## Testing
>
>
> New/updated test coverage:
> - `tests/core/serve-command.spec.ts` — cases over `withServeAddress`: npm via `--`, pnpm/yarn/bun direct, custom host flag, `--prefix` scripts, trimming, the fallback command, plus table-driven cases for non-forwarding shapes (`vite`, `next dev`, chained commands, already-pinned ports) and for injection-shaped strings.
> - `tests/cli/site_dev.spec.ts` — default sandbox address, config-supplied address, rejection of `--port`, refusal of a non-forwarding `serveCommand`, serving without a login, the `--hostname` spelling, same-origin env (`URL=undefined`), and the no-`site`-block failure. Driven by `runLive` against `serve.js` fixtures that echo their argv and env.
> - `tests/cli/install.spec.ts` — configured `installCommand` runs, works without a login, surfaces a failing command, fails without a site block.
> - `tests/cli/build.spec.ts` — builds without a logged-in user.
> - `tests/core/project.spec.ts` — the three new schema defaults.
> - New fixtures: `with-dev-address`, `with-hostname-serve-command`, `with-npm-serve-command`, `with-installable-site`, `with-failing-install`.
>
> ## Checklist
>
> - [ ] My changes generate no new warnings
>
> ## Additional Notes
>
> Division of labour between the three dev paths: `base44 dev` is the developer machine (local backend), `base44 dev --remote` is the developer machine against the published backend, and `base44 site dev` is the hosted-sandbox path (frontend only, same-origin backend). `base44 dev` is unaffected by the new config keys — it binds nothing.
>
> The commit history walks through several designs for where the bind address lives (command flags, then `--backend-url`, then config); the final shape is config-only, so `base44 site dev` needs no arguments and no caller has to know the values.
>
> `docs/` topic guides were not touched — the new commands follow the existing factory / `Base44Command` pattern rather than changing it — but the new `site.dev*` config keys may be worth a mention in the project-structure docs.
>
> ---
> <sub>🤖 Generated by Claude | 2026-09-23 13:04 UTC | c3a4ce4</sub>


packages/cli/src/cli/program.ts
```diff
@@ -77,6 +78,7 @@ export function createProgram(context: CLIContext): Command {
   program.addCommand(getCreateCommand());
   program.addCommand(getScaffoldCommand());
   program.addCommand(getDashboardCommand());
+  program.addCommand(getInstallCommand());
```
netanelgilad: let's place install under `base44 site` as well cause its under site in the config and its only relevant if there is a site

packages/cli/src/cli/commands/site/dev.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (!site) {
+    throw new InvalidInputError(
+      "This project has no 'site' block in base44/config.jsonc, so there is no frontend to serve. Add one naming its serveCommand; site dev falls back to \"npm run dev\".",
+    );
+  }
+
+  // Serving is this command's whole job, so an unnamed dev server is a
+  // convention rather than "nothing to run". Everything else comes from the
+  // config, which is why this command takes no arguments for it.
+  const serveCommand = site.serveCommand ?? DEFAULT_SERVE_COMMAND;
```
netanelgilad: serveCommand has a default at the schema level so no need for the fallback here

packages/cli/src/cli/commands/site/dev.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!site) {
+    throw new InvalidInputError(
+      "This project has no 'site' block in base44/config.jsonc, so there is no frontend to serve. Add one naming its serveCommand; site dev falls back to \"npm run dev\".",
+    );
+  }
+
+  // Run as written: where the dev server binds is the command's own business.
+  // In a sandbox @base44/vite-plugin binds 0.0.0.0:5173 for Base44 apps; any
+  // other serveCommand must bind the address the sandbox exposes itself.
+  const command = site.serveCommand ?? DEFAULT_SERVE_COMMAND;
```
netanelgilad, line 37: the defualt is coming from the schema, so there is no need for another default here?

netanelgilad (approved):

packages/cli/src/cli/program.ts
ronnyrin: Done in f4434ef — it's `base44 site install` now, beside `site deploy` and `site dev`. No top-level alias kept: nothing outside this PR called it.

packages/cli/src/cli/commands/site/dev.ts
ronnyrin: `serveCommand` is deliberately not defaulted in the schema. It's `z.string().optional()`, unlike `buildCommand` and `installCommand` right beside it. `base44 dev` reads its absence as "this project has no frontend" and runs the backend alone (`dev.ts:65`). We tried defaulting it in #634 and CI failed: `full-project` has a site block with no `serveCommand`, `base44 dev` spawned `npm run dev`, the dev server exited, and its exit handler shut the backend down with it. So the fallback lives in `site dev`, whose only job is serving, where an unnamed command can only mean `npm run dev`.

packages/cli/src/cli/commands/site/dev.ts
ronnyrin: Same question as the newer thread on this line; answered there: https://github.com/base44/cli/pull/635#discussion_r4091209252

ozsay (approved):

